PR #11, opened 2021-04-20
Update readme and example for usage of TF registry

### LINKED ISSUE #9: Update examples & README to v1
Refs #7

The README and `examples/` both reference 0.2 as the latest version. These should be updated for v1. I think the sources need to also be updated to point to the registry.

jpcoenen:
Updated versions to `~> 1.0.0` and replaced the GitHub reference with the TF registry name in all but the local usage example.

Resolves #9.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,6 +2,25 @@
 
 Use the 1Password Connect Terraform Provider to reference, create, or update items in your 1Password Vaults.
 
+## Usage
+
+```tf
+terraform {
+  required_providers {
+    onepassword = {
+      source = "1Password/onepassword"
+      version = "~> 1.0.0"
+    }
+  }
+}
+
+provider "onepassword" {
+  url     = "http://<1Password Connect API Hostname>"
+}
+```
+
+See the [examples](./examples/) directory for a full example. 
+
 ## Building
 
 To build the 1Password Connect Terraform provider run the following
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -2,6 +2,25 @@
...
+  required_providers {
+    onepassword = {
+      source = "1Password/onepassword"
+      version = "~> 1.0.0"
+    }
+  }
+}
+
+provider "onepassword" {
+  url     = "http://<1Password Connect API Hostname>"
```
florisvdg, line 18:
```suggestion
  url = "http://<1Password Connect API Hostname>"
```

florisvdg (commented):
Better, but if we call it `Usage` then I'd expect a little `onepassword_item` example to be in there as well.

README.md
```diff
@@ -2,6 +2,43 @@
...
+  }
+}
+
+provider "onepassword" {
+  url = "http://localhost:8080"
+}
+
+variable "demo_vault" {
+  default = "fuacthv5i7gextkqv45g7algw8"
+}
```
florisvdg, line 23: May be cleaner to just do:
```suggestion
variable "demo_vault" {}
```

florisvdg:
Still some `demo_vault` occurrences in the examples.

florisvdg (approved):

Issue #15, opened 2021-04-29
Tags don't get updated to 1password vault


bondezbond:
### Your environment

Terraform Provider Version: 1.0.0
Connect Server Version: 1.0.0
OS: macOS Big Sur
Terraform Version: 0.15.1

## What happened?
I'm using the example project provided looking at the code. I see `tags` tags a list(string). I added `tags = ["test"]` to all resources in example project. The apply shows the tags being added but when you look at the state file I see tags being an empty list but apply completed successfully. The resource in vault also don't show any tags either. When I run apply again it tries to add the tags again.

## What did you expect to happen?
If I pass a list of tags it successfully adds the tags 

## Steps to reproduce
1. Add tags to all resources in example project. Tried with one tag or two tags
2. Run plan & apply

```
resource "onepassword_item" "demo_password" {
  vault = var.demo_vault

  title    = "Demo Password Recipe"
  category = "password"

  password_recipe {
    length  = 40
    symbols = false
  }

  tags = ["test"]
}

resource "onepassword_item" "demo_login" {
  vault = var.demo_vault

  title    = "Demo Terraform Login"
  category = "login"
  username = "test@example.com"
  tags = ["test"]
}

resource "onepassword_item" "demo_db" {
  vault    = var.demo_vault
  category = "database"
  type     = "mysql"

  title    = "Demo TF Database"
  username = "root"

  database = "Example MySQL Instance"
  hostname = "localhost"
  port     = 3306
  tags = ["test"]
}

resource "onepassword_item" "demo_sections" {
  vault = var.demo_vault

  title    = "Demo Terraform Item with Sections"
  category = "login"
  username = "test@example.com"

  section {
    label = "Terraform Section"

    field {
      label = "API_KEY"
      type  = "CONCEALED"
      value = "2Federate2!"
    }

    field {
      label = "HOSTNAME"
      value = "example.com"
    }
  }

  section {
    label = "Terraform Second Section"

    field {
      label = "App Specific Password"
      type  = "CONCEALED"

      password_recipe {
        length  = 40
        symbols = false
      }
    }

    field {
      label = "User"
      value = "demo"
    }
  }
  tags = ["test"]
}
```

## Notes & Logs
```
2021-04-29T12:51:43.814-0400 [DEBUG] provider.terraform-provider-onepassword_v1.0.0: 2021/04/29 12:51:43 [ERROR] setting state: Invalid address to set: []string{"notesPlain"}
[...]

Issue #13, opened 2021-04-29
When changing item category generated password gets lost


bondezbond:
### Your environment

Terraform Provider Version: 1.0.0
Connect Server Version: 1.0.0

OS: macOs Big Sur 11.2.3

Terraform Version: 0.15.2

## What happened?
Created a test op item as `category: password` with only password recipe all other variables blank. It create the password item successfully and I see the password in vault. When changing the category to `login` terraform shows that it'll change the category type BUT it blanks out the password in the vault. Example code

```
resource "onepassword_item" "item" {
  vault = var.vault

  title    = var.name // test
  category = var.category // password -> login

  username = var.username // default ""
  database = var.database // default ""
  hostname = var.hostname // default ""
  port     = var.port // default ""

  password_recipe {
    length  = 40
    symbols = false
  }
}
```
![image](https://user-images.githubusercontent.com/4400773/116544542-4aabf200-a8bd-11eb-9765-f5c668229438.png)

## What did you expect to happen?
I expected it to change the category type successfully and the resource in vault would convert the item to login without it blanking out the password in vault.

If changing category types aren't possible I would expect the provider to error out.

## Steps to reproduce
1. Run apply with category as `password`
2. Run second apply with category as `login`

jpcoenen:
Hi :wave: 

Thank you for your extensive bug report! You are very right that changing the category should not be possible in the way it is currently allowed. I opened a [PR](https://github.com/1Password/terraform-provider-onepassword/pull/16) to make sure that Terraform recreates the item when you change the category.

PR #19, opened 2021-05-04
Release v1.0.1

jpcoenen:
## Fixes
- Tags set in the `tags` field now correctly get set on items in your vault. {#15}
- Changing the category of an item no longer results in the contents of your item no longer being visible. {#13}
- Changing the vault of an item no longer leads to an error that the item cannot be found.


thatguygriff (approved):

bondezbond:
@jpcoenen This release doesn't seem to be available yet on the terraform registry

jpcoenen:
@bondezbond Thanks making me aware of that. The release was not yet public; it is now and should be available on the Terraform Registry.

PR #23, opened 2021-05-14
Add GitHub Action to check if TF docs are up-to-date

jpcoenen:
This Action will fail if the current docs to not correspond to the output of `go generate`.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,31 @@
+name: Test
+on: [push, pull_request]
+
+jobs:
+  verify-docs:
+    name: Verify docs
+    runs-on: ubuntu-latest
+    steps:
+      - name: Set up Go 1.x
+        uses: actions/setup-go@v2
+        with:
+          go-version: ^1.15
+
+      - name: Install Terraform
+        uses: hashicorp/setup-terraform@v1
+
+      - name: Check out code into the Go module directory
+        uses: actions/checkout@v2
+
+      - name: Generate docs
+        run: go generate
+
+      - name: Check if git detects no changes
+        run: |
+          if output=$(git status --porcelain) && [ -z "$output" ]; then
+            echo "Docs are up to date";
+          else
+            echo "Not all docs are up-to-date";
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -0,0 +1,31 @@
+name: Test
```
SimonBarendse, line 1: We also have a [`Build & Test` workflow](https://github.com/1Password/terraform-provider-onepassword/blob/c3fe693801bae9140fb46af1399ff780f59b2912/.github/workflows/build.yml#L1).

I believe naming this workflow `Test` may create the expectations that we run the go test suite in this workflow.

How about naming this workflow `Verify Docs` instead?

SimonBarendse (approved):
Minor comment, non-blocking

florisvdg:
Why just validate while you could also commit while you're at it?

.github/workflows/test.yml
```diff
@@ -0,0 +1,31 @@
...
+
+      - name: Check if git detects no changes
+        run: |
+          if output=$(git status --porcelain) && [ -z "$output" ]; then
+            echo "Docs are up to date";
+          else
+            echo "Not all docs are up-to-date. Run 'go generate' and commit the changes.";
+            echo $output;
+            exit 1;
+          fi
```
florisvdg, line 31: I think this should do it:
```suggestion
      - name: Commit go generate if needed
        run: |
          git config user.name "GitHub Actions"
          git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
          git add .
          git commit -m "Go generate"
          git push
```

thatguygriff: Big ++ on this one :)

jpcoenen: Cool idea! But I am hesitant with adding an automated commit without it being properly tested.

As it's not needed to reach the goal of this PR, I am going to merge the PR as-is for now. Feel free to open a separate PR for this addition.

thatguygriff (approved):

PR #29, opened 2021-06-07
Clean up readme to focus on usage

SimonBarendse:
This should make it easier for users who just want to use the provider to find the right information. Previously, the instructions for local installation may have let them to believe that was necessary for usage, while it is only necessary to locally develop the provider.

Users who want to contribute to the provider, can now find detailed instructions in a separate doc.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@ Use the 1Password Connect Terraform Provider to reference, create, or update ite
 
 ## Usage
 
-Detailed documentation for using this provider can be found [here](https://registry.terraform.io/providers/1Password/onepassword/latest/docs).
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -4,6 +4,8 @@ Use the 1Password Connect Terraform Provider to reference, create, or update ite
 
 ## Usage
 
+Detailed documentation for using this provider can be found [here](https://registry.terraform.io/providers/1Password/onepassword/latest/docs).
```
mackenbach, line 7:
```suggestion
Detailed documentation for using this provider can be found on the [Terraform Registry docs](https://registry.terraform.io/providers/1Password/onepassword/latest/docs).
```

mackenbach (approved):
Nice, one small tweak.

florisvdg (approved):

PR #36, opened 2021-06-24
Use Go 1.16 during release job

jpcoenen:
This makes sure the darwin/arm64 binary is also included.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -18,7 +18,7 @@ jobs:
         name: Set up Go
         uses: actions/setup-go@v2
         with:
-          go-version: 1.15
+          go-version: 1.16
       -
         name: Import GPG key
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -37,4 +37,4 @@ jobs:
           args: release --rm-dist
         env:
           GPG_FINGERPRINT: ${{ steps.import_gpg.outputs.fingerprint }}
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
\ No newline at end of file
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
jpcoenen, line 40: Apparently `vim` also fixed the missing newline.

verkaufer (approved):

PR #45, opened 2021-09-27
Fix docs regarding the vault data source

edif2008:
Two small adjustments in the docs:
 - the vault data source has the `uuid` read-only parameter
 - the `id` of the item data source has been clarified


jpcoenen (approved):
Way clearer :+1:

edif2008:
Looks good to me.
The docs are clearer now.

Issue #46, opened 2021-10-04
"Missing required attribute" error (URL)


nnovaes:
trying the terraform integration using a pretty "standard" config for 1password on terraform and I'm getting this error:

**│ Error: Missing required attribute
│ 
│ on line 1:
│ (source code not available)
│ 
│ The attribute "url" is required, but no definition was found.
**
my docker compose file is:
```
version: "3.4"

services:
  op-connect-api:
    image: 1password/connect-api:latest
    ports:
      - "8080:8080"
    volumes:
      - "./1password-credentials.json:/home/opuser/.op/1password-credentials.json"
      - "data:/home/opuser/.op/data"
  op-connect-sync:
    image: 1password/connect-sync:latest
    ports:
      - "8081:8080"
    volumes:
      - "./1password-credentials.json:/home/opuser/.op/1password-credentials.json"
      - "data:/home/opuser/.op/data"

volumes:
  data:

```

and my terraform file is

```
terraform {
  required_providers {
    docker = {
      source  = "kreuzwerker/docker"
      version = "2.15.0"
    }
    onepassword = {
      source = "1Password/onepassword"
      version = "~> 1.1.2"

    }
  }
}

provider "onepassword" {
  url = "http://localhost:8080"
}

resource "onepassword_item" "demo_password" {
  vault = "ACTUAL_VAULT_UUID"

  title    = "Demo Password Recipe"
  category = "password"

  password_recipe {
    length  = 40
    symbols = false
  }
}

```

output of versions

```
❯ terraform version
Terraform v1.0.8
on darwin_amd64
+ provider registry.terraform.io/1password/onepassword v1.1.2
+ provider registry.terraform.io/kreuzwerker/docker v2.15.0
❯ docker version
Client:
 Cloud integration: 1.0.17
 Version:           20.10.7
 API version:       1.41
 Go version:        go1.16.4
 Git commit:        f0df350
 Built:             Wed Jun  2 11:56:22 2021
 OS/Arch:           darwin/amd64
 Context:           default
 Experimental:      true

Server: Docker Engine - Community
 Engine:
  Version:          20.10.7
  API version:      1.41 (minimum version 1.12)
  Go version:       go1.13.15
  Git commit:       b0f5bc3
  Built:            Wed Jun  2 11:54:58 2021
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.4.6
[...]

PR #47, opened 2021-10-04
Fix providing token through prompt

jpcoenen:
As reported in https://github.com/1Password/terraform-provider-onepassword/issues/46, Terraform can throw the following error when only the `url` attibrute is set in the provider configuration block:

```
The attribute "url" is required, but no definition was found.
```

This happens when the `token` attribute is provided through the prompt instead of by setting the `OP_CONNECT_TOKEN` environment variable and is caused by Terraform's weird handling of multiple required attributes for a provider.

I suggest to address this by turning the `url` field in an optional field for Terraform and verifying that it is actually set in the `ConfigureFunc`. In practice this means that `token` and `url` are still both required, but that only `token` will be prompted for. It also resolves Terraform's weird behaviour of forgetting the `url` when set.

### DIFF
--- a/onepassword/provider.go
+++ b/onepassword/provider.go
@@ -36,7 +36,8 @@ func Provider() *schema.Provider {
 			"url": {
 				Type:        schema.TypeString,
 				Required:    true,
-				Description: "The HTTP(S) URL where your 1Password Connect API can be found",
+				DefaultFunc: schema.EnvDefaultFunc("OP_CONNECT_HOST", nil),
+				Description: "The HTTP(S) URL where your 1Password Connect API can be found. Can also be sourced from OP_CONNECT_HOST.",
 			},
 			"token": {
 				Type:        schema.TypeString,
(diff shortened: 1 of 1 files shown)


docs/index.md
```diff
@@ -22,4 +22,7 @@ provider "onepassword" {
 ### Required
 
 - **token** (String, Required) A valid token for your 1Password Connect API. Can also be sourced from OP_CONNECT_TOKEN.
-- **url** (String, Required) The HTTP(S) URL where your 1Password Connect API can be found
+
+### Optional
+
+- **url** (String, Optional) The HTTP(S) URL where your 1Password Connect API can be found. Can also be sourced from OP_CONNECT_HOST.
```
florisvdg, line 28: This is a bit confusing. It's not optional. Could use some clarification.

jpcoenen: What about
```suggestion
### Optional

- **url** (String, Optional) The HTTP(S) URL where your 1Password Connect API can be found. Must be provided through the the OP_CONNECT_HOST environment if this attribute is not set.
```

florisvdg: Yes, much better.

Issue #54, opened 2022-01-19
Cannot create (T)OTP fields


drdaeman:
### Your environment

Terraform Provider Version: 1.1.3

Connect Server Version: 1.5.0

OS: any

Terraform Version: 1.0.7, but it shouldn't matter

## What happened?
Trying to create a field with type "TOTP" causes a validation error coming from the Connect server:

    Error: status 400: Validation: (validateVaultItem failed to Validate), Couldn't validate the item: "[ItemValidator] has found 1 errors, 0 warnings: \nErrors: {1. details.sections[1].fields[0] has unsupported field type: totp}"

According to the https://github.com/1Password/connect/pull/26 the correct value for `field.type` is "OTP" rather than "TOTP". Can you please fix the `fieldTypes` array accordingly?

https://github.com/1Password/terraform-provider-onepassword/blob/906b16d5c09d924acfe20301bc34499c0938cdcc/onepassword/resource_onepassword_item.go#L56

Thank you!

## Steps to reproduce

    resource "onepassword_item" "something" {
      ...
      section {
        ...
        field {
          label = "One-time password"
          type = "TOTP"
          value = ...
        }
      }
    }

edif2008:
Good news. We have just released v1.1.4 which contains the fix to (T)OTP field type. 🎉 
Thank you for reporting this bug and also providing a PR with the fix! 😄

PR #55, opened 2022-01-19
Fix (T)OTP field type

drdaeman:
As I've reported in #54, the correct `field.type` is `"OTP"` rather than `"TOTP"`. In this PR I've made relevant changes in hope that it would save you some time editing files :)

I've tested this using a local provider override in my infrastructure against 1Password Connect 1.5.0 server and was successfully able to create a bunch of items with OTP fields, so I can vouch this change actually works.

I was not able to update existing items, changing existing `field`'s `type` from `"CONCEALED"` to `"OTP"` - `terraform apply` had the correct plan and succeeded but nothing had actually changed. However, I believe that's a separate issue that probably lies with the Connect servers rather than this Terraform provider (although it can be argued that provider could be told that such change requires re-creating of a resource). Destroying those items and re-creating them anew worked.

Hope this could get merged and a new version of provider released, so we would be able to use OTP fields soon!

Thanks!


jpcoenen:
Thank you for addressing this! 🙌

jpcoenen (approved):

PR #62, opened 2022-06-09
Update github.com/hashicorp/go-getter to v1.6.1

jpcoenen:
This addresses a few vulnerabilities in the dependency.

These vulnerabilities should have no impact on this Terraform Provider.


kpcraig:
It seems good to me - all tests pass for a dep-only update

PR #63, opened 2022-06-14
OP CLI Basic Support

smartinov:


gaffneyd4:
Conceptually, I would much rather use the already-authenticated op cli on my local workstation to interact with the 1password api over having to set up and configure the connect server. The main use case I had in mind that directed me to look for a 1password terraform provider in the first place is just being able to source (read only) sensitive variables from my 1password vaults instead of having to set them as environment variables on my system. In the meantime I'm going to look into what is available for running op cli commands in a script to populate variables in tf. I think something like [this](https://stackoverflow.com/questions/47311916/how-to-put-in-varable-tf-to-run-script-with-data-external-in-terraform) would work.

Tensho:
It was a surprise that I couldn't use my user auth material and had to issue a personal service account (SA) to run Terraform from a local machine. I understand the case when you need SA in CI or any other automated system, but human run should be run with a result of the local authentication (`op signin`). Wonder what is the root of this design decision?

While this PR is incomplete it looks like it does the right things.

volodymyrZotov:
This was implemented in #79. @smartinov Thank you very much for that idea 👍

Authentification as a user is added in #119.

Issue #59, opened 2022-04-07
Section and field labels' case is not preserved.


midu-git:
### Your environment

Terraform Provider Version: 1.1.4

Connect Server Version: 1.5.1

OS: macOS

Terraform Version:1.1.7

## What happened?
Define an item in 1password and assign a new section to it, "DATA" say. Inside this section, define a field, "appName" say. When calling this item using onepassword_item, the sections' label is "Data" and the label of the mentioned field is "appname".

## What did you expect to happen?
Sections' label should be "DATA" and fields' label should be "appName", i.e. the case should be preserved.
At least, both labels should be consistently modified => "data", "appname".

## Steps to reproduce
As described above.

## Notes & Logs
For the fields' name, the provider explicitly converts it to lower case
```go
dataField["label"] = strings.ToLower(f.Label)
```
The label of the section, however, is not modified in the provider. It is returned from a call to the sdk
```
github.com/1Password/connect-sdk-go
```
and there, just a 
```go
json.Unmarshal(body, result)
```
is called. Thus, the conversion needs to happen either in connect, or returned from 1password API, directly.

midu-git:
Is anyone outside there noticing these bug reports?

midu-git:
@edif2008 What does it mean that you labeled this bug? From outside it seems, that there is no further process here. Can you comment on this or even provide a timeline? Is there someone working on this issue?

edif2008:
Hey @midu-git,

Apologies for the late reply. The `tracked` label means that we're monitoring this issue in our internal platforms. This is done to make sure that this issue doesn't fall through the cracks.

As for a timeline, I unfortunately cannot give you anything concrete. I will bump this issue with my team in the hopes that someone has the time to pick it up.

midu-git:
Hey @edif2008. Thank you the clarification.

PR #69, opened 2022-08-04
do not lowercase item label

volodymyrZotov:
Resolves: #59 

*Description:*
Item's label returned as lowercase value from provider.

*Solution:*
Return item's label as it is returned from connect.

### DIFF
--- a/onepassword/data_source_onepassword_item.go
+++ b/onepassword/data_source_onepassword_item.go
@@ -185,7 +185,7 @@ func dataSourceOnepasswordItemRead(data *schema.ResourceData, meta interface{}) 
 			if f.Section != nil && f.Section.ID == s.ID {
 				dataField := map[string]interface{}{}
 				dataField["id"] = f.ID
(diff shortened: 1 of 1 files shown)


kpcraig:
this seems nice - do we know why we were lowercasing in the first place?

kpcraig (commented):
Do we need to change the casing on [line 211](https://github.com/1Password/terraform-provider-onepassword/blob/4d2da3ca6cd6e44276002d05403dd76f579339c5/onepassword/data_source_onepassword_item.go#L211)?

volodymyrZotov:
> Do we need to change the casing on [line 211](https://github.com/1Password/terraform-provider-onepassword/blob/4d2da3ca6cd6e44276002d05403dd76f579339c5/onepassword/data_source_onepassword_item.go#L211)?

Looks like we don't as key and value are strings.

jillianwilson (approved):

Issue #79, opened 2023-05-08
Feature Request: Service Account support


gabrielrinaldi:
### Summary
Is there a plan to add service account support to Terraform?

### Use cases
I have a private cluster and connection to 1Password would be a lot easier through a SA instead of having to deploy a Terraform Agent in the cluster

### Proposed solution
Add SA support to the Terraform provider

### Is there a workaround to accomplish this today?
Today I would have to deploy a Terraform Agent in the cluster and install the 1Password Operator to give access to secrets

### References & Prior Work
* GitHub Actions can use SA

dustin-ruetz:
Hi @gabrielrinaldi - thanks for reaching out with this feature request. I can confirm that using Service Accounts with Terraform is on our radar, but I can't provide any commitments or timelines as of this writing. Thank you again for expressing your interest!

gabrielrinaldi:
Thanks for getting back to me @dustin-ruetz, I think that would be game changing in replacing Vault (I was using that prior) as it would allow for more security and simplify password management incredibly. Please keep me posted as I am super interested in this.

gabrielrinaldi:
@dustin-ruetz is this something I could help with? I don't know how to build `terraform` providers, but I am willing to try as secrets are a little out of control given that I can't use them in Terraform unless I open `connect` to the internet.

franklouwers:
@dustin-ruetz Hi, would this still be on the short-term radar? I have a few customers where setting up and hosting a separated Connect wouldn't make a lot of sense, but using an SA would be the perfect solution.

TakumiHaruta:
This is really good. We're eagerly awaiting this feature!

ekostjuk:
Hacky ugly workaround:
Improved something that was previously shared on 1p slack by another dev: https://gist.github.com/ekostjuk/944d3c4483c1961fee805b66eb1679ac

xophere:
Yeah this provider is basically pointless without service account support.

xophere:
Yeah we just opted to use a different tool because the template process isn't gonna cut it.

gabrielrinaldi:
@xophere what vendor are you using? I am evaluating AWS Secrets and Vault Open Source as options this does not seem ready for production yet.

xophere:
[...]

PR #99, opened 2023-09-11
Add Service Account support

tim-oster:
This PR adds support for using an 1Password Service Account token as opposed to using a Connect Server setup (#79).

The implementation is best effort but most likely not ideal. It is a wrapper around the `op` CLI to be able to make use of the service account token. As of today, there seems to be no other option to interface with 1password other than the cli or a connect server.

The biggest issue in this implementation is, that updating an item will first delete it and then createa new one. The main reason for this is, that `op` does not support updating items using json. Ideally, we would use `FroceNew` for this in the provider but since it needs to be compatible with the Connect Server mode, this is not possible.


franklouwers:
Does this mean that when updating an item, it will get a new UUID and new urls?

tim-oster:
@franklouwers yes, unfortunately.

It might be tweakable to use the actual `op item edit` command. The only downside I see is that it might not be fully compatible with everything that the json schema offers and that it would essentially leak all credentials of the command as arguments ([more info](https://developer.1password.com/docs/cli/reference/management-commands/item/#item-edit-flags)).

gabrielrinaldi:
@tim-oster does that mean the agent running this needs the `op` cli installed? We are using Terraform Cloud, which I think would not support that unfortunately. But excited to see progress on support for SAs.

volodymyrZotov:
@tim-oster Great contribution! Your work will help us a lot to release this feature even faster! 🚀
I'll review your pr shortly.

onepassword/cli/op.go
```diff
@@ -0,0 +1,109 @@
...
+}
+
+func (op *OP) UpdateItem(ctx context.Context, item *onepassword.Item, vaultUuid string) (*onepassword.Item, error) {
+	if item.Vault.ID != "" && item.Vault.ID != vaultUuid {
+		return nil, errors.New("item payload contains vault id that does not match vault uuid")
+	}
+	item.Vault.ID = vaultUuid
+
+	// op cli does not support easy updating of an item by passing in a json payload
+	err := op.DeleteItem(ctx, item, vaultUuid)
```
volodymyrZotov, line 87: I'd not go with deleting/creating the item as you'll need to operate with new `uuid` each time you re-created it and keeping in mind Service Accounts [quotas and rate limits](https://developer.1password.com/docs/service-accounts/get-started#rate-limits).
But I see the reason you implemented it in this way.

I'd like to have the ability to simply update the existing one and use a similar implementation as for  `ItemCreate`. I'll raise that within the team to see what we can do with that.

onepassword/cli/exec.go
```diff
@@ -0,0 +1,86 @@
...
+
+func parseCliError(stderr []byte) error {
+	subMatches := cliErrorRegex.FindStringSubmatch(string(stderr))
+	if len(subMatches) != 3 {
+		return fmt.Errorf("unkown op error: %s", string(stderr))
+	}
+	return fmt.Errorf("op error: %s", subMatches[2])
+}
+
+func (op *OP) exec(ctx context.Context, dst any, stdin []byte, args ...opArg) error {
```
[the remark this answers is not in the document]
tim-oster: The distinction was made to in attempt to keep the internal workings clutter out of the public interface defition, especially with all the flag wrappers and so on. But I see your point, refactored.

onepassword/cli/op.go
tim-oster: Totally agree, I don't like the way it is implemented either at atm. If we could get a json edit flow, that would be amazing.

tim-oster:
@volodymyrZotov thanks for the review!

@gabrielrinaldi yes, unfortunately using the CLI seems to be the only option at the moment. I am not too familiar with tf cloud, do you have access to a local file system? In that case, it might be worth investigating if we can bundle the cli binary together with the provider binary. Though this would drastically increase the provider's binary size.

volodymyrZotov:
@gabrielrinaldi [Only Install Standalone Binaries](https://developer.hashicorp.com/terraform/cloud-docs/run/install-software#only-install-standalone-binaries) article might be helpful to understand how you can add additional binaries.

And this one [Install 1Password CLI](https://developer.1password.com/docs/cli/get-started/#step-1-install-1password-cli) might help to install. See `Linux -> apt`, or whatever is suitable for you.

Hope those will help you to figure out how to do that  😃 Let me know how it's going for you, please.

gabrielrinaldi:
@volodymyrZotov amazing, thanks for pointing it out! Can't wait for this to be merged.

volodymyrZotov:
Hi folks! And especially @tim-oster 👋 Just want to put an update about the item edit. 

Implementing that with the current cli version brings some difficulties, so **we decided to add a piped input feature to the cli `item edit` command**, which would allow us to pass the `Item` struct to perform the updates. So with the newer cli version you will be able to implement `EditItem` function in the same way as `CreateItem`.

I'm on that feature right now. I'll keep you posted about the progress. The initial aim is to have a new cli release with this feature later next week, but will see how it goes. 🚀

tim-oster:
@volodymyrZotov amazing, thanks for having a look at this so quickly! I'll update the PR as soon the CLI is released. Looking forward to getting this merged 💪

volodymyrZotov:
Hi @tim-oster. Just an update on that. We're still working on the **item edit piped input feature**. I'll let you know as soon as we have a new cli release with it and happy to help bring your PR to the finish line 🙂

maxpowis-bp:
I just started a complex implementation an realize I might just be stuck at this stage due to this bug. Any ETA for the release ?

speier:
In the meantime here is a hack using `op` cli locally..

```hcl
variable "item_ref" {}

locals {
  temp_dir = "${path.root}/op_tmp"
}

resource "local_file" "output" {
  filename = "${local.temp_dir}/${timestamp()}.txt"
  content  = ""
}

resource "null_resource" "op_read" {
  triggers = {
    once = timestamp()
  }
  provisioner "local-exec" {
    command = "op read ${var.item_ref} > ${local_file.output.filename}"
  }
}

data "local_file" "output" {
  depends_on = [null_resource.op_read]
  filename   = local_file.output.filename
}

resource "null_resource" "deleteLocalFile" {
  depends_on = [null_resource.op_read, data.local_file.output]
  triggers = {
    once = timestamp()
  }
  provisioner "local-exec" {
    command = "rm -rf ${local.temp_dir}"
  }
}

output "item_value" {
  value = data.local_file.output.content
}
```

usage example:

```hcl
module "op" {
  source   = "../modules/op"
  item_ref = "op://myvault/fooprovider/api-token"
}

output "result" {
  value = module.op.item_value
}
```

maxpowis-bp:
Thanks for the quick answer, I am using terraform-cloud though; probably won't do the trick :/

hollow:
Having to fall back to calling `op` from Terraform either through `local-exec` or from the provider code sounds like the wrong direction for this feature and will prevent Terraform Cloud users from using it.

speier:
Sure, totally agree guys, `local-exec` is not the way. It's just a temp hack. Ideally a proper provider should work with 1Password API.

volodymyrZotov:
@tim-oster Hi. Some updates about **`op item edit` piped input feature**.
I know that your PR is blocked by that, but want to let you know that we're actively working on that feature on our side. 
I'll drop a message here once it's released. Stay tuned. 😃

volodymyrZotov:
@tim-oster Hi! 👋 We plan to release a new CLI version that includes the feature to edit items using templates later this week. 
Would you be able to continue the work on the PR? 😃

dbrennand:
@tim-oster @volodymyrZotov - Looks like the new CLI version was released 🙂 https://app-updates.agilebits.com/product_history/CLI2#v2230001

tim-oster:
@volodymyrZotov @dbrennand the new CLI update is awesome! Works as expected!

dbrennand:
> @volodymyrZotov @dbrennand the new CLI update is awesome! Works as expected!

Awesome 🙌🏻 - hopefully this gets merged soon! 😃

volodymyrZotov:
@tim-oster Hi. I pushed some commits to slightly change the error messages and address a couple of issues (see comments above).

While testing I found at least one issue that we need to address before moving forward. It can't create passwords when using CLI and throws an error:
```
[ERROR] 2023/11/21 17:32:33 unable to process line 1: Validation: (validateVaultItem failed to Validate), Couldn't validate the item: "[ItemValidator] has found 1 errors, 0 warnings: \nErrors:{1. Password item requires ps value}"
```

I'll continue looking into that tomorrow and continue testing to see if I can find any other issues.

volodymyrZotov (approved):

volodymyrZotov:
Added myself as a co-author to sign all the commits.

Issue #96, opened 2023-07-18
1.2.0 use of sections does not work


ralphkootker:
When using version 1.2.0 the usage of sections stopped working as expected compared to version 1.1.2 till 1.1.4:

`
resource "onepassword_item" "workspace_details" {
  vault = var.vault_uuid

  title    = var.name
  category = "login" # Only supports login, database of notes

  password = var.token
  #username = "not-required"

  section {
    label = "Workspace"
    field {
      label = "workspace_id"
      type  = "STRING"
      value = var.workspace_id
    }
    field {
      label = "workspace_url"
      type  = "STRING"
      value = var.workspace_url
    }
  }
}
`

buginventor:
Maybe this info will help with investigation.
In the state there is no `id` for `field` and that's why it's not showing in GUI (app) nor UI (web).
it looks like:
```
user@blabla dir % terraform state show onepassword_item.some_configuration
# onepassword_item.some_configuration:
resource "onepassword_item" "some_configuration" {
    category = "password"
    id       = "vaults/xxx/items/yyy"
    password = (sensitive value)
    tags     = [
        "some",
        "tags",
    ]
    title    = "Something xxx"
    uuid     = "xxx"
    vault    = "xxx"

    section {
        id    = "1-2-2-3-4"
        label = "Label x"

        field {
            id      = ""
            label   = "XML"
            type    = "CONCEALED"
            value   = (sensitive value)
        }
    }
}


```

gogetdane:
+1 this. We've downgraded to 1.1.4. Terraform reports resources as existing, yet in the UI and CLI resources do not contain sections.

If I add a new section in TF, the apply output shows the expected sections although none exist on the item in 1P.  We've pinned to 1.1.4, tainted affected resources and recreated.

col:
We're having the same issue. This is a pretty significant bug! I'm disappointed this is not getting more attention. 😞

JBG-brad:
With the time gap between releases and the lack of attention to major bugs I'm curious if the 1Password has abandoned this project.

volodymyrZotov:
Hi @JBG-brad. This one is currently on our radar and we'll try to address it as soon as possible. I understand how frustrating it is. Sorry for the delay.

PR #100, opened 2023-10-17
Fix item creation with sections

williamhpark:
## Overview

Currently, creating items with sections through the Terraform provider does not work.

It seems like the `Purpose` field for item section fields is being assigned improperly. If a `type` is specified when creating a resource, this `type` value is being assigned to the section field's `Purpose`. This causes an error, since the expected values for `Type` and `Purpose` differ. For more details on the expected values for each, refer to the documentation here: https://developer.1password.com/docs/connect/connect-api-reference/#item-field-object.

I also made a small change to the logic for generating a section ID, to make it cleaner.

## Type of change

Bug fix

## Related Issue(s)

- Resolves: #96 

## How To Test

1. Pull the branch locally: `git pull && git checkout wpark/96-fix-sections`
2. Configure a provider override in your `~/.terraformrc` file, so that you use a local build of the provider rather than the published version. Refer to the documentation here for more details: https://developer.hashicorp.com/terraform/cli/config/config-file#development-overrides-for-provider-developers. Your `.terraformrc` file should look like this:

```
provider_installation {

  dev_overrides {
      "1Password/onepassword" = "<path to terraform-provider-onepassword repo root directory>"
  }

  # For all other providers, install them directly from their origin provider
  # registries as normal. If you omit this, Terraform will _only_ use
  # the dev_overrides block, and so no other providers will be available.
  direct {}
}
```

3. Navigate to the root of the `terraform-provider-onepassword` repo and generate a local build: `go build`
4. Create an item resource with sections using `terraform apply`. Follow the instructions [here](https://github.com/1Password/terraform-provider-onepassword/tree/main/examples) - the example `main.tf` contains a `demo_sections` resource that you can use to test this

 - [ ] In the vault that your connect server has access rights to, the item should have been successfully created with sections, and the item should match the resource outlined in `main.tf`. If you used the example provided in the repo, the item should look like this:
 
image

5. Check the generated item within Terraform state by running `terraform state show`, e.g. `terraform state show onepassword_item.demo_sections`

- [ ] The outputted resource should match the resource outlined in `main.tf`, and the `id` for all sections and fields should be populated

### DIFF
--- a/onepassword/resource_onepassword_item.go
+++ b/onepassword/resource_onepassword_item.go
@@ -546,19 +546,17 @@ func dataToItem(data *schema.ResourceData) (*onepassword.Item, error) {
 		if !ok {
 			return nil, fmt.Errorf("Unable to parse section: %v", sections[i])
 		}
-		sid, err := uuid.GenerateUUID()
-		if err != nil {
-			return nil, fmt.Errorf("Unable to generate a section id: %w", err)
-		}
 
-		if section["id"].(string) != "" {
-			sid = section["id"].(string)
-		} else {
+		if section["id"].(string) == "" {
+			sid, err := uuid.GenerateUUID()
+			if err != nil {
+				return nil, fmt.Errorf("Unable to generate a section id: %w", err)
+			}
 			section["id"] = sid
 		}
 
 		s := &onepassword.ItemSection{
-			ID:    sid,
+			ID:    section["id"].(string),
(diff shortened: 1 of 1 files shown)


volodymyrZotov (approved):

volodymyrZotov:
✅ The code looks good! Tested and it works as expected!

PR #121, opened 2023-12-11
Update documentation

### LINKED ISSUE #115: Provide the suggested way how to securely setup provider in the documentation.
Currently, in the documentation we provide either this example to set up provider
```
// README.md
provider "onepassword" {
  url = "http://localhost:8080"
}
```
or this
```
// README in Terraform Registry (`templates/index.md.tmpl` file)
provider "onepassword" {
  URL                                 = "http://localhost:8080"
  token                               = "CONNECT_TOKEN"
  service_account_token = "SERVICE_ACCOUNT_TOKEN"
  op_cli_path                     = "OP_CLI_PATH"
}
```

Setting the sensitive values (tokens) directly in the terraform script is not secure, as it can be accidentally pushed and potentially leak the secrets. Therefore, the suggested and more secure way to set up the provider will be to use env variables `OP_SERVICE_ACCOUNT_TOKEN` to use service accounts and `OP_CONNECT_TOKEN` to use Connect.

We need to update the documentation to reflect this.

volodymyrZotov:
Mention service account in the examples to encourage users to use them. 
Provide examples of how to set up the provider using env variables and mention that's the more secure way to do so.
Mention minimum CLI `v2.23.0`.

Resolves #115


docs/index.md
```diff
@@ -60,13 +65,22 @@ provider "onepassword" {
...
-- `url` (String) The HTTP(S) URL where your 1Password Connect API can be found. Must be provided through the OP_CONNECT_HOST environment variable if this attribute is not set. Must be set to use with 1Password Connect server.
+- `account` (String) A valid account's sign-in address or ID to use biometrics unlock. Can also be sourced from OP_ACCOUNT environment variable. Provider will use 1Password CLI if set.
+- `op_cli_path` (String) The path to the 1Password CLI binary. Can also be sourced from OP_CLI_PATH environment variable. Defaults to `op`.
+- `service_account_token` (String) A valid 1Password service account token. Can also be sourced from OP_SERVICE_ACCOUNT_TOKEN environment variable. Provider will use 1Password CLI if set.
+- `token` (String) A valid token for your 1Password Connect API. Can also be sourced from OP_CONNECT_TOKEN environment variable. Provider will use 1Password Connect server if set.
+- `url` (String) The HTTP(S) URL where your 1Password Connect API can be found. Can also be sourced OP_CONNECT_HOST environment variable. Provider will use 1Password Connect server if set.
```
edif2008, line 86: `[nitpick]`
Environment variable names should be wrapped with backticks (`` ` ``) for better highlighting. would it be possible to do that when generating this part of the documentation?

Also, usually for the Connect related configuration we present first the parameter for the URL, followed by the one for the token. Is it possible to make this ordering when generating this part of the documentation?

```suggestion
- `account` (String) A valid account's sign-in address or ID to use biometrics unlock. Can also be sourced from `OP_ACCOUNT` environment variable. Provider will use 1Password CLI if set.
- `op_cli_path` (String) The path to the 1Password CLI binary. Can also be sourced from `OP_CLI_PATH` environment variable. Defaults to `op`.
- `service_account_token` (String) A valid 1Password service account token. Can also be sourced from `OP_SERVICE_ACCOUNT_TOKEN` environment variable. Provider will use 1Password CLI if set.
- `url` (String) The HTTP(S) URL where your 1Password Connect API can be found. Can also be sourced `OP_CONNECT_HOST` environment variable. Provider will use 1Password Connect server if set.
- `token` (String) A valid token for your 1Password Connect API. Can also be sourced from `OP_CONNECT_TOKEN `environment variable. Provider will use 1Password Connect server if set.
```

templates/index.md.tmpl
```diff
@@ -17,6 +17,8 @@ You must install [1Password CLI](https://developer.1password.com/docs/cli) on th
 
 ## Use with 1Password CLI
 
+Minimum 1Password CLI version `v2.23.0` is required.
```
edif2008, line 20: `[non-blocking]`

Should this be highlighted more with emojis or using bold? I worry that it may be easily missed, and it's an important thing to call out.

README.md
```diff
@@ -27,7 +27,7 @@ terraform {
 }
 
 provider "onepassword" {
-  url = "http://localhost:8080"
+  service_account_token = "<1Password service account token>"
```
edif2008, line 30: Are we sure this is sufficient in terms of ensuring users will provide the service account token in a secure way? 🤔

onepassword/provider.go
```diff
@@ -44,31 +44,31 @@ func Provider() *schema.Provider {
...
 				DefaultFunc: schema.EnvDefaultFunc("OP_CONNECT_TOKEN", nil),
-				Description: "A valid token for your 1Password Connect API. Can also be sourced from OP_CONNECT_TOKEN. Must be set to use with 1Password Connect server.",
+				Description: "A valid token for your 1Password Connect API. Can also be sourced from OP_CONNECT_TOKEN environment variable. Provider will use 1Password Connect server if set.",
 			},
 			"service_account_token": {
 				Type:        schema.TypeString,
 				Optional:    true,
 				DefaultFunc: schema.EnvDefaultFunc("OP_SERVICE_ACCOUNT_TOKEN", nil),
-				Description: "A valid token for your 1Password Service Account. Can also be sourced from OP_SERVICE_ACCOUNT_TOKEN. Must be set to use with 1Password service account.",
+				Description: "A valid 1Password service account token. Can also be sourced from OP_SERVICE_ACCOUNT_TOKEN environment variable. Provider will use 1Password CLI if set.",
```
edif2008, line 59: `[nitpick]`
```suggestion
				Description: "A valid 1Password service account token. Can also be sourced from OP_SERVICE_ACCOUNT_TOKEN environment variable. Provider will use the 1Password CLI if set.",
```

onepassword/provider.go
```diff
@@ -44,31 +44,31 @@ func Provider() *schema.Provider {
...
 			},
 			"account": {
 				Type:        schema.TypeString,
 				Optional:    true,
 				DefaultFunc: schema.EnvDefaultFunc("OP_ACCOUNT", nil),
-				Description: "A valid account's sign-in address or ID to use biometrics unlock. Can also be sourced from OP_ACCOUNT. Must be set to use with biometric unlock.",
+				Description: "A valid account's sign-in address or ID to use biometrics unlock. Can also be sourced from OP_ACCOUNT environment variable. Provider will use 1Password CLI if set.",
```
edif2008, line 65: `[nitpick]`
```suggestion
				Description: "A valid account's sign-in address or ID to use biometrics unlock. Can also be sourced from OP_ACCOUNT environment variable. Provider will use the 1Password CLI if set.",
```

templates/index.md.tmpl
```diff
@@ -39,17 +43,27 @@ URL                        EMAIL                                         USER ID
...
-2. Set `url` and `token` in the provider configuration.
+2. Set `OP_CONNECT_HOST` environment variable or `url` in the provider configuration.
+3. Set `OP_CONNECT_TOKEN` environment variable or `token` in the provider configuration.
 
 ## Example Usage
 
 {{tffile "examples/provider/provider.tf"}}
 
+Each of the configuration options can be set using env variables.
+That's more secure way to do that as prevents from accidentally pushing sensitive values to the project repo.
```
edif2008, line 61: `[nitpick]`
Better English phrasing:
```suggestion
Each of the configuration options can be set using env variables.
This is a more secure way to do provide your credentials to avoid accidentally pushing sensitive values to the project repository.
```

CONTRIBUTING.md
```diff
@@ -42,7 +42,7 @@ terraform {
 }
 
 provider "onepassword" {
-  url = "http://<1Password Connect API Hostname>"
+  service_account_token = "<1Password service account token>"
```
edif2008, line 45: `[question]`
While this promotes using the provider with a service account, we should mention the other possible ways of configuring the provider: either as minimized code examples or link to the relevant documentation.

The reason for this is because the contributors may not use service accounts in their flow, but rather one of the other configurations and it would help to share them here to have a smooth onboarding for any type of configuration.

edif2008 (approved):
## Code review: ✅ 

The documentation improvements look good and makes things clearer and more secure.

I have a couple of notes:
- `[maybe blocking]` The Terraform files in the `examples` directory contain **only** Connect configurations. I assume it would help to provide some examples in which we configure the provider to use either service accounts or regular user (upcoming feature). Would adding these examples be part of this PR? Or should we add them in a separate one?
- `[question]`  I've re-read the expected syntax for `op_cli_path` (or `OP_CLI_PATH`) and it expects the _entire_ path to the `op` binary? For some reason I thought that it points to the directory where the binary lives (e.g. `/usr/local/bin/` rather than `/usr/local/bin/op`). It may be out of scope of this PR, but I would like to get some clarifications about this parameter.
-  `[question]` In the README we now promote the flow using service accounts, which is great. What is a bit concerning is that now people can easily miss the other configuration options: using Connect or authenticating as a regular user. We should either point to the proper documentation about this or provide the examples with these configuration as minimized code snippers (i.e. using `<details></details>` blocks).

CONTRIBUTING.md
armstrongl: +1

docs/index.md
```diff
@@ -60,13 +65,22 @@ provider "onepassword" {
 }
 ```
 
+Each of the configuration options can be set using env variables.
+That's more secure way to do that as prevents from accidentally pushing sensitive values to the project repo.
```
armstrongl, line 69:
```suggestion
Environment variables are a more secure way to set configuration options because they prevent accidentally pushing sensitive values to the project repo.
```

docs/index.md
```diff
@@ -60,13 +65,22 @@ provider "onepassword" {
 }
 ```
 
+Each of the configuration options can be set using env variables.
```
armstrongl, line 68:
```suggestion
You can set each of the configuration options using environment variables.
```

docs/index.md
```diff
@@ -60,13 +65,22 @@ provider "onepassword" {
...
+That's more secure way to do that as prevents from accidentally pushing sensitive values to the project repo.
+
+- `OP_CONNECT_HOST` - The URL where your 1Password Connect API can be found.
```
armstrongl, line 71: In the Connect documentation, we typically refer to this environment variable as the Connect server's URL/hostname/IP (not the Connect API)... I know it's because we're accessing the API, but I think we should keep it consistent 🙂

armstrongl (approved):
This adds a lot of valuable context. I didn't dig in too deeply because I'm going to migrate a lot of the Terraform documentation to the Developer Portal soon. I'll do a more thorough review at that point.

volodymyrZotov:
@armstrongl Thank you very much for taking a look!

README.md
volodymyrZotov: The idea of having `service_accouunt_token` here is to copy/paste the snippet and run it.

More details on how to use more secure way to setting env vars are in [Terraform Registry docs](docs/index.md) and [examples](examples/README.md)

docs/index.md
volodymyrZotov: Regarding backticks - yes, we can wrap env vars.

Regarding the order - no, it's generated automatically in alphabetical order.

templates/index.md.tmpl
volodymyrZotov: added emoji

CONTRIBUTING.md
volodymyrZotov: More details on how to setup the provider users can find in [Terraform Registry docs](https://github.com/1Password/terraform-provider-onepassword/pull/docs/index.md) and [examples](https://github.com/1Password/terraform-provider-onepassword/pull/examples/README.md)

The docs above are the source of true and entry point. The other examples show one of the possible ways of setting it up.
The benefit of that would be to reduce the amount of docs we'll need to support which makes it clean. Also users have to read less 😄

Let me know if that makes sense.

edif2008: I'm not suggesting to copy-paste the documentation for the other configuration options. Adding a link to them like this should suffice:

> Check [this documentation](inset-link-here) for more details about provider configuration.


This doesn't increase a lot the size of the documentation, while also ensuring that the user knows where to reach other documentation (rather than us expecting them to look for the place where it is e.g.  [Terraform Registry docs](https://github.com/1Password/terraform-provider-onepassword/pull/docs/index.md)). This improves the user journey in terms of learning how to use the tool to be a great contributor. We shouldn't assume that they know where to find the info. The last advantage of this (which can be nitpicky) is that this results of the user making one click to reach to the right documentation rather than....at least 3-5 with the other approach (i.e. getting to the docs on their own).

Anyway, we can always iterate on this, so I'm fine with it as it is now. 😄

PR #127, opened 2023-12-14
fix: fix op cli always generating passwords

### LINKED ISSUE #128: Terraform cannot create items with the password we provide in the code
### Your environment

Terraform Provider Version:1password provider version:1.3.1

Connect Server Version:No connect server.Service account token is used.

CLI Version:2.23.0

OS:Windows

Terraform Version:1.6.5

## What happened?
 I am trying to add new passwords into 1password. I want the items to have the password that I mention in the code through environment variables. But when I apply the terraform changes, a random password is created, which is not expected.

## What did you expect to happen?
I want the password in the code to appear in 1password.

## Steps to reproduce
1. Create an item in 1password using terraform code: example snippet:
resource "onepassword_item" "secrets-1"{
  vault = var.vault
  title    = "test"
  category = "password"
  password = "sdfyguihgfg"
}

2)When we do terraform apply, a different password is created other than the one mentioned in the code snippet.

josh-burton:
When invoking the op cli, if the onepassword item had no password recipe, generate-password would always be appended to the cli args and any password value would be overwritten by a generated password.

Note I am very new to Go so please point me in the right direction if needed :)

As part of this fix I have extracted the logic which constructs the generate-password cli arg into a function in utils, and implemented a test around it.

Fixes #128

### DIFF
--- a/onepassword/cli/op.go
+++ b/onepassword/cli/op.go
@@ -100,12 +100,9 @@ func (op *OP) create(ctx context.Context, item *onepassword.Item, vaultUuid stri
 	args := []opArg{p("item"), p("create"), p("-")}
 	// 'op item create' command doesn't support generating passwords when using templates
 	// therefore need to use --generate-password flag to set it
-	if pf := passwordField(item); pf != nil {
-		recipeStr := "letters,digits,32"
-		if pf.Recipe != nil {
-			recipeStr = passwordRecipeToString(pf.Recipe)
-		}
-		args = append(args, f("generate-password", recipeStr))
+	recipe := passwordRecipe(item)
+	if recipe != "" {
+		args = append(args, f("generate-password", recipe))
 	}
 
 	err = op.execJson(ctx, &res, payload, args...)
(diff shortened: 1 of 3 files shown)


edif2008 (approved):
Thank you so much for providing a fix for the bug mentioned in #128. I really appreciate it. 💪🏻 

## Functional review: ✅ 

Validated that this code will no longer generate passwords when creating an item and the user provided a custom value to the password.

## Code review: ✅ 

Code looks clean and straight forward. I really appreciate you extracting the functionality for getting the password recipe for an item, as well as writing tests for that new functionality. 

I left one nitpick that would simplify some code the code, however the current implementation is good already.

## Other notes

Could you add in the PR description that it resolves issue #128? In this way, when we merge this PR, the issue will close automatically. 😄 

Adding this line should suffice:
```md
Resolves: #128 
```

josh-burton:
@edif2008 thanks for the review :) I've updated the PR description and applied your suggestion.

PR #153, opened 2024-02-27
Pass proper user agent info to the CLI

### LINKED ISSUE #124: Set proper values for 'OP_INTEGRATION_*' env vars
We need to set proper 'OP_INTEGRATION_*' env var values when using CLI.

Currently, they are:
```
cmd.Env = append(cmd.Environ(),
		"OP_FORMAT=json",
		"OP_INTEGRATION_NAME=terraform-provider-connect",
		"OP_INTEGRATION_ID=GO",
		//"OP_INTEGRATION_BUILDNUMBER="+version.ProviderVersion, // causes bad request errors from CLI
	)
```

It should be changed to:
```
cmd.Env = append(cmd.Environ(),
		"OP_FORMAT=json",
		"OP_INTEGRATION_NAME=terraform-provider",
		"OP_INTEGRATION_ID=TP",
		"OP_INTEGRATION_BUILDNUMBER="+version.ProviderVersion,
	)
```

Note:
Setting `OP_INTEGRATION_BUILDNUMBER` env var caused CLI error. In scope of this issue, we need to investigate how to pass to properly so the CLI throws no errors.

edif2008:
This PR brings the user-agent info passed to the CLI up-to-date.

Resolves #124.

Specifically, the user agent info is as follows:

| Environment variable | Value |
|:--------------------:|:------:|
| `OP_INTEGRATION_NAME` | `terraform-provider` |
| `OP_INTEGRATION_ID` |  `TFP` |
| `OP_INTEGRATION_BUILDNUMBER` | build version of the provider e.g. `1040201` |

**Notes:** 
- I've changed the ID compared to the issue from `TP` to `TFP` to match the 3-letter pattern that we use for the ID in other integrations.
- The `makeBuildVersion` function is identical to the one we've done for the [`kubernetes-secrets-injector`](https://github.com/1Password/kubernetes-secrets-injector/blob/main/pkg/utils/build_version.go).

### DIFF
--- a/onepassword/cli/op.go
+++ b/onepassword/cli/op.go
@@ -220,9 +220,9 @@ func (op *OP) execRaw(ctx context.Context, stdin []byte, args ...opArg) ([]byte,
 	cmd := exec.CommandContext(ctx, op.binaryPath, cmdArgs...)
 	cmd.Env = append(cmd.Environ(),
 		"OP_FORMAT=json",
-		"OP_INTEGRATION_NAME=terraform-provider-connect",
-		"OP_INTEGRATION_ID=GO",
 		//"OP_INTEGRATION_BUILDNUMBER="+version.ProviderVersion, // causes bad request errors from CLI
+		"OP_INTEGRATION_NAME=terraform-provider",
+		"OP_INTEGRATION_ID=TFP",
 	)
 	if op.serviceAccountToken != "" {
 		cmd.Env = append(cmd.Env, "OP_SERVICE_ACCOUNT_TOKEN="+op.serviceAccountToken)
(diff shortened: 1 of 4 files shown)


volodymyrZotov:
@edif2008 I tested it, but this version doesn't with those env vars set. The CLI returns the following error.
```
Error: op error: error initializing client: Validation: (400) (Bad Request), The structure of the request was invalid.
```
I can confirm if I come back to the version in the `main` branch, it works.

edif2008:
Very curious that it doesn't work. We do the same thing in the injector, so I wonder what's different.

I've also tried to set the environment to the CLI as is (so not through the provider) and it seems that it doesn't like any build number, which is very weird.

edif2008:
Found the issue: The server doesn't recognize this new integration type. Will make an internal issue to fix this.

volodymyrZotov (approved):

PR #161, opened 2024-03-20
Add support for document data source

sdahlbac:
I have have only tested this with unittests and cli, as I do not have a connect.

- Supports multiple files
- Supports text documents and binary documents (provided one uses the `content_base64` version)

Roundtrip test:

```hcl
data "onepassword_item" "foo" {
  vault = var.demo_vault
  uuid  = "..."
}
resource "local_file" "truststore" {
  filename       = "truststore.jks"
  content_base64 = data.onepassword_item.foo.file.2.content_base64
}
```
(where 2 is the third file for that item, added via the UI)

I had to use the `op read` approach rather than `op document get` (unless I am missing an undocumented feature) to be able to handle multiple files.

I have not tried to implement a resource, since as far as I understood connect cannot support it and I don't see a way to distinguish between the two backends.

~~The `document` category is available also for resource, that is slightly incorrect.~~

### DIFF
--- a/onepassword/data_source_onepassword_item.go
+++ b/onepassword/data_source_onepassword_item.go
@@ -2,6 +2,7 @@ package onepassword
 
 import (
 	"context"
+	"encoding/base64"
 	"errors"
 	"fmt"
 	"strings"
@@ -154,6 +155,39 @@ func dataSourceOnepasswordItem() *schema.Resource {
 					},
 				},
 			},
+			"file": {
+				Description: sectionFilesDescription,
+				Type:        schema.TypeList,
+				Computed:    true,
+				MinItems:    0,
+				Elem: &schema.Resource{
+					Description: fileDescription,
+					Schema: map[string]*schema.Schema{
+						"id": {
+							Description: fileIDDescription,
+							Type:        schema.TypeString,
+							Computed:    true,
+						},
+						"name": {
+							Description: fileNameDescription,
+							Type:        schema.TypeString,
+							Computed:    true,
+						},
+						"content": {
+							Description: fileContentDescription,
+							Type:        schema.TypeString,
+							Computed:    true,
+							Sensitive:   true,
+						},
+						"content_base64": {
+							Description: fileContentBase64Description,
+							Type:        schema.TypeString,
+							Computed:    true,
+							Sensitive:   true,
+						},
+					},
+				},
+			},
 		},
 	}
 }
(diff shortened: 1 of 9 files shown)


onepassword/data_source_onepassword_item.go
```diff
@@ -177,6 +211,22 @@ func dataSourceOnepasswordItemRead(ctx context.Context, data *schema.ResourceDat
 		}
 	}
 
+	dataFiles := []interface{}{}
```
volodymyrZotov, line 214: Could you please create struct instead of interface and describe all the properties there? Might be something like `FileSchema`

onepassword/data_source_onepassword_item_test.go
```diff
@@ -126,6 +154,25 @@ func compareItemToSource(t *testing.T, dataSourceData *schema.ResourceData, item
...
+		if err != nil {
+			t.Errorf("Unexpected error occured")
+		}
+		if dataSourceData.Get(fmt.Sprintf("file.%d.content", i)).(string) != string(want) {
+			t.Errorf("Expected file %v to have content %v, got %v", file.Name, string(want), dataSourceData.Get(fmt.Sprintf("file.%d.content", i)))
+		}
+	}
+	if len(item.Files) == 2 {
+		t.Errorf("dafuq")
+	}
```
volodymyrZotov, line 175: I think this can be removed

onepassword/resource_onepassword_item.go
```diff
@@ -38,13 +38,21 @@ const (
 	sectionLabelDescription  = "The label for the section."
 	sectionFieldsDescription = "A list of custom fields in the section."
 
+	sectionFilesDescription = "A list of files in an item."
```
volodymyrZotov, line 41:
```suggestion
	filesListDescription = "A list of files in an item."
```

onepassword/data_source_onepassword_item.go
sdahlbac: I did that, but it seems to be a bit convoluted since we need to have the properties with terraform naming in the ResourceData, but the struct needs to be public for the data to come across (at least that is what I saw when fixing...)

also, there is prior art wrt `[]interface{}{}` in the datasource.

But less dynamism is probably a good thing after all.

sdahlbac:
Ran `go generate` to get updated docs, but it seems that `terraform-plugin-docs` fails to include descriptions of nested fields 😢

sdahlbac:
@volodymyrZotov comments adressed

sdahlbac:
Speculating with myself: Actually, the fact that descriptions are missing from the schema might be a slight issue if things like `sensitive: true` is also missing

sdahlbac:
It seems that the cheese has been moved. I'll see if I can figure out what changes are needed due to the change of tf framework.

Also, it seems that #164 is somewhat related.

sdahlbac:
Ok, I have something that seems to work locally against main. Will create a superseding PR when I have cleaned up it a little bit more.

sdahlbac:
Superceded by #171 

@volodymyrZotov would you review that one instead, please?

volodymyrZotov:
@sadiaazmal Hi, was about to write to you that there was a migration and you would need to adjust your PR because of that, but I see you figured it out already 👍 Will take another look there.

Issue #132, opened 2024-01-13
Feature Request: Item Files


bloudraak:
### Summary
The ability to create, delete, read and list files attached to an item

### Use cases
As part of automation, we obtain certificates from a certificate vendor. However, we need to store the private key in a secure location. The certificate vendor cannot access the private key at any point. We also need to keep the public key and certificate chain. We often store that information on the host (if the certificates are tied to a specific host) or in its own item when the domain is broad.

Later on, when we provision infrastructure, we need to retrieve the certificates to add them to hosts etc securely. 

### Proposed solution

Introduce a new data source and resource:

```
data "onepassword_file" "this" {
    item = "<itemuuid>"
    name = "private.key"
}

resource "local_file" "demo" {
  content = jsonencode(onepassword_file.this.content)
  filename = "demo.json"
}

resource "onepassword_file" "that" {
    item = "<itemuuid2>"
    field = "field1"
    content = file("private.key")
}

```

Also introduce a data source to retrieve all files attached to an item, which returns its metadata (name etc), but not the contents

```
data "onepassword_files" "this" {
    item = "<itemuuid>"
}

# gets the contents of each file on the item
data "onepassword_file" "this" {
   for_each = data. onepassword_files.this.files
   item = each.value.item
   field = each.value.id
}

```

The reason this is separate is because their lifecycle may be different than that of the item they belong to. For example, the public key of a certificate or a license file for a given host may change every year, while the item they belong to remains untouched.

### Is there a workaround to accomplish this today?

* Using the CLI.  This is problematic when running on CI/CD infrastructure as it's an additional requirement.
* Convert the file to base64 and store it as a field. This can be problematic since the lifecycle of the field is directly tied to the lifecycle of the item. (see #133 to address this)

### References & Prior Work

For the "pattern" see the following:

Data Sources:
* [aws_s3_object](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/s3_object)
* [aws_s3_objects](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/s3_objects)
* [content_library_item](https://registry.terraform.io/providers/hashicorp/vsphere/latest/docs/data-sources/content_library_item)

Resources:
* [s3_object](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/s3_object)
* [content_library_item](https://registry.terraform.io/providers/hashicorp/vsphere/latest/docs/resources/content_library_item)

Implementation:
[...]

PR #164, opened 2024-05-01
Feature: File attachments to items

asgeirn:
Adds `files` to the item resource, with properties to get the file name and contents.

Uses `op read` to get the file contents after transforming the content path of the file to an `op://` URL.

I tried scanning through the `op` documentation for other ways of doing this, but was unable to find a way to reliably use for instance `op document get`

Fixes 1Password/terraform-provider-onepassword#132


volodymyrZotov:
Implemented in #161

Issue #51, opened 2021-11-09
Item resource to support files and documents when using Service Account or desktop auth via SDK only


clark42:
### Summary
Add the ability to retrieve (with a new or existing data in terraform) and store (with a new or existing terraform resources) files (document category).

### Use cases
We save somes privates and secure document in our vault, like ssh private key or other certificate.
We would like to retrieve them with the terraform provider to avoid retrieving those files on our filesystem.

### Proposed solution
From this discussion on your community forum (https://1password.community/discussion/120671/secrets-automation-1password-documents), it seems that the connect-api now support file https://support.1password.com/connect-api-reference/#get-file-content since a few months (August 2021)

Don't hesitate to ask more details if needed!

edif2008:
Hey @clark42,

Thank you for raising this. I can see the use of Document type Items with the Terraform provider.
Currently the Connect API only allows to download Document Items and Items that contain files attached to them. The API endpoint for uploading files with Connect is something that is on our backlog. You can follow [this issue](https://github.com/1Password/connect/issues/11) to find out when the API endpoint is released.

We have added this feature into our backlog and will keep you updated when we release a new version of the provider with support for getting Document type Items.

andrew8er:
Any updates on this? It's been quiet some time…

volodymyrZotov:
Item data source for `Document` category is supported since [v2.1.0](https://github.com/1Password/terraform-provider-onepassword/releases/tag/v2.1.0).

Uploading documents is still blocked by lack of this feature on [1Password Connect side](https://github.com/1Password/connect/issues/11). I'll find out more about that feature on Connect side and drop an update.

volodymyrZotov:
Currently, provider can be used in 2 ways:
- With 1Password Connect (using Connect credentials)
- With 1Password SDK (using Service Account or desktop authentication (biometric unlock))

There is no plans to add support for files creation/update/delete on Connect side. At least in a short term.

But SDK supports [managing files](https://developer.1password.com/docs/sdks/files).

We would be happy to make the provider partially support Documents/Files for item resource only when using Service Account or desktop authentication (via SDK).

Though, this feature it's not in our roadmap yet, the team would appreciate the help of community in building it and would be happy to support this effort with giving a feedback and PR reviews.

PR #171, opened 2024-05-17
feat/document on plugin framework

sdahlbac:
This is a re-implementation of #161 due to the upgrading to the new plugin framework.

I have have only tested this with acceptance and cli, as I do not have a connect.

- Supports multiple files
- Supports text documents and binary documents (provided one uses the content_base64 version)
Roundtrip test:

data "onepassword_item" "foo" {
  vault = var.demo_vault
  uuid  = "..."
}
resource "local_file" "truststore" {
  filename       = "truststore.jks"
  content_base64 = data.onepassword_item.foo.file.2.content_base64
}
(where 2 is the third file for that item, added via the UI)

I had to use the op read approach rather than op document get (unless I am missing an undocumented feature) to be able to handle multiple files.

I have not tried to implement a resource, since as far as I understood connect cannot support it. I added a validation that you cannot create a new `onepassword_item` with category `document`. But the error message does not say why you cannot do it.

(While it sucks a bit having to reimplement this out of the blue, I must say that the first impressions of the newer framework feels nicer to work with.)

Relates: #51

### DIFF
--- a/docs/resources/item.md
+++ b/docs/resources/item.md
@@ -66,7 +66,7 @@ resource "onepassword_item" "demo_db" {
 
 ### Optional
 
-- `category` (String) The category of the item. One of ["login" "password" "database" "secure_note"]
+- `category` (String) The category of the item. One of ["login" "password" "database" "secure_note" "document"]
 - `database` (String) (Only applies to the database category) The name of the database.
 - `hostname` (String) (Only applies to the database category) The address where the database can be found
 - `note_value` (String, Sensitive) Secure Note value.
--- a/docs/resources/item.md
+++ b/docs/resources/item.md
@@ -66,7 +66,7 @@ resource "onepassword_item" "demo_db" {
 
 ### Optional
 
-- `category` (String) The category of the item. One of ["login" "password" "database" "secure_note" "document"]
+- `category` (String) The category of the item. One of ["login" "password" "database" "secure_note"]
 - `database` (String) (Only applies to the database category) The name of the database.
 - `hostname` (String) (Only applies to the database category) The address where the database can be found
 - `note_value` (String, Sensitive) Secure Note value.
--- a/internal/onepassword/cli/op.go
+++ b/internal/onepassword/cli/op.go
@@ -208,6 +208,20 @@ func (op *OP) delete(ctx context.Context, item *onepassword.Item, vaultUuid stri
 	return nil, op.execJson(ctx, nil, nil, p("item"), p("delete"), p(item.ID), f("vault", vaultUuid))
 }
 
+func (op *OP) GetFileContent(ctx context.Context, file *onepassword.File, itemUuid, vaultUuid string) ([]byte, error) {
+	versionErr := op.checkCliVersion(ctx)
+	if versionErr != nil {
+		return nil, versionErr
+	}
+	ref := fmt.Sprintf("op://%s/%s/%s", vaultUuid, itemUuid, file.ID)
+	tflog.Debug(ctx, "reading file content from: "+ref)
+	res, err := op.execRaw(ctx, nil, p("read"), p(ref))
+	if err != nil {
+		return nil, err
+	}
+	return res, nil
+}
+
 func (op *OP) execJson(ctx context.Context, dst any, stdin []byte, args ...opArg) error {
 	result, err := op.execRaw(ctx, stdin, args...)
 	if err != nil {
--- a/internal/provider/const.go
+++ b/internal/provider/const.go
@@ -30,6 +30,13 @@ const (
 	sectionLabelDescription  = "The label for the section."
 	sectionFieldsDescription = "A list of custom fields in the section."
 
+	filesDescription             = "A list of files attached to the item."
+	fileDescription              = "A file attached to the item."
+	fileIDDescription            = "A UUID for the file."
+	fileNameDescription          = "The name of the file."
+	fileContentDescription       = "The content of the file."
+	fileContentBase64Description = "The content of the file in base64 encoding. (Use this for binary files.)"
+
 	fieldDescription        = "A custom field."
 	fieldIDDescription      = "A unique identifier for the field."
 	fieldLabelDescription   = "The label for the field."
@@ -57,6 +64,7 @@ var (
 		strings.ToLower(string(op.Password)),
 		strings.ToLower(string(op.Database)),
 		strings.ToLower(string(op.SecureNote)),
+		strings.ToLower(string(op.Document)),
 	}
 
 	fieldPurposes = []string{
--- a/internal/provider/const.go
+++ b/internal/provider/const.go
@@ -29,10 +29,11 @@ const (
 	sectionIDDescription     = "A unique identifier for the section."
 	sectionLabelDescription  = "The label for the section."
 	sectionFieldsDescription = "A list of custom fields in the section."
+	sectionFilesDescription  = "A list of files attached to the section."
 
 	filesDescription             = "A list of files attached to the item."
 	fileDescription              = "A file attached to the item."
-	fileIDDescription            = "A UUID for the file."
+	fileIDDescription            = "The UUID of the file."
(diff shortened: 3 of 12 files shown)


internal/provider/test_http_server.go
```diff
@@ -53,6 +65,20 @@ func setupTestServer(expectedItem *onepassword.Item, expectedVault onepassword.V
 				if err != nil {
 					t.Errorf("error writing body: %s", err)
 				}
+			} else if r.URL.String() == fmt.Sprintf("/v1/vaults/%s/items/%s/files/%s/content", expectedItem.Vault.ID, expectedItem.ID, files[0].ID) ||
+				r.URL.String() == fmt.Sprintf("/v1/vaults/%s/items/%s/files/%s/content", expectedItem.Vault.ID, expectedItem.ID, files[1].ID) {
```
sdahlbac, line 69: is feels like there should be a smarter way of handling multiple similar paths

docs/resources/item.md
```diff
@@ -66,7 +66,7 @@ resource "onepassword_item" "demo_db" {
 
 ### Optional
 
-- `category` (String) The category of the item. One of ["login" "password" "database" "secure_note"]
+- `category` (String) The category of the item. One of ["login" "password" "database" "secure_note" "document"]
```
edif2008, line 69: This documentation is inaccurate and can lead to confusion for customers. 😕

internal/provider/onepassword_item_resource_test.go
```diff
@@ -155,13 +156,34 @@ func TestAccItemResourceWithSections(t *testing.T) {
 	})
 }
 
+func TestAccItemResourceDocument(t *testing.T) {
```
edif2008, line 159: It would be nice to add another test case in which we get an item of a different category with item attachments to it in different sections, since this is a possibility.

internal/provider/test_http_server.go
edif2008: An idea that might help here is to use regex:
```go
filePath := regexp.MustCompile("/v1/vaults/[a-z0-9]*/items/[a-z0-9]*/files/[a-z0-9]*/content")
```

and then use it in the conditional like so:
```go
} else if filePath.MatchString(r.URL.String()) {
```

internal/onepassword/cli/op.go
```diff
@@ -208,6 +208,20 @@ func (op *OP) delete(ctx context.Context, item *onepassword.Item, vaultUuid stri
 	return nil, op.execJson(ctx, nil, nil, p("item"), p("delete"), p(item.ID), f("vault", vaultUuid))
 }
 
+func (op *OP) GetFileContent(ctx context.Context, file *onepassword.File, itemUuid, vaultUuid string) ([]byte, error) {
```
edif2008, line 211: `[nitpick]`

This will make the CLI client part of the Terraform provider slightly less performant, since `GetItem` (as in `op item get`) already gets the content of the files the item has. So now what will happen is to fetch the item twice instead of once.

internal/provider/onepassword_item_data_source.go
```diff
@@ -298,6 +332,19 @@ func (d *OnePasswordItemDataSource) Read(ctx context.Context, req datasource.Rea
 		}
 	}
 
+	for _, f := range item.Files {
+		content, err := d.client.GetFileContent(ctx, f, item.ID, item.Vault.ID)
```
edif2008, line 336: `[nitpick]`

What might be useful here is to call the function if the file's content doesn't exist yet, which is the case for Connect, but not for the CLI.

internal/provider/const.go
```diff
@@ -30,6 +30,13 @@ const (
 	sectionLabelDescription  = "The label for the section."
 	sectionFieldsDescription = "A list of custom fields in the section."
 
+	filesDescription             = "A list of files attached to the item."
+	fileDescription              = "A file attached to the item."
+	fileIDDescription            = "A UUID for the file."
```
edif2008, line 35: `[nitpick]`

To match the same styling of the rest of documentation:
```suggestion
	fileIDDescription            = "The UUID of the file."
```

edif2008 (changes requested):
## Summary

Thank you so much for your effort in making this contribution. This is much appreciated and it looks pretty solid.

## Functional review: ✅ 

I've validated the following scenarios and can confirm that they work as expected:
- A Document item with a file attached to it
- An item of different category (e.g. Login) with multiple files attached to it.

One thing that would be nice to do here is to put the files in their appropriate section, since this is a possibility through the 1Password app. I don't think it's possible via the CLI though.

I also ran the acceptance tests and they all pass.

## Code review: ✅ 

The code is straight forward and easy to follow. I've left comments in how to make it better.

The thing to call out code-wise is related to file attachments being able to belong to sections as well. This is something that might require some changes to ensure we include this scenario as well.

internal/onepassword/cli/op.go
sdahlbac: I actually tried this, and when running tests everything looked good, but when I actually tried using the provider (with op cli) i got the content not yet loaded error, so it did not work

internal/provider/onepassword_item_data_source.go
sdahlbac: fixed

docs/resources/item.md
sdahlbac: fixed

internal/provider/onepassword_item_resource_test.go
```diff
@@ -155,13 +156,34 @@ func TestAccItemResourceWithSections(t *testing.T) {
...
 	return fmt.Sprintf(`
-
-data "onepassword_vault" "acceptance-tests" {
-	uuid = "%s"
-}	
-resource "onepassword_item" "test-database" {
+	data "onepassword_vault" "acceptance-tests" {
+		uuid = "%s"
+	}
+	resource "onepassword_item" "test-database" {
```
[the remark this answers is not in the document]
sdahlbac: fixed

internal/provider/onepassword_item_data_source.go
```diff
@@ -187,6 +196,31 @@ func (d *OnePasswordItemDataSource) Schema(ctx context.Context, req datasource.S
 					},
 				},
 			},
+			"file": schema.ListNestedBlock{
```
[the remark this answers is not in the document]
sdahlbac: implemented section support

sdahlbac:
@edif2008 The review comments should be adressed now

edif2008 (approved):
## Functional review: ✅ 

Tested and validated that files from sections are placed in the appropriate sections in the Terraform state.

## Code review: ✅ 

Code looks much cleaner now. Thank you for addressing the feedback.

## Other notes

⚠️ It seems that the documentation check fails since the `/docs/data-sources/item.md` file is not properly updated. Re-running `go generate` and pushing the changes to that file should fix it.

Once that is addressed, this PR is good to be merged. 😄

sdahlbac:
@edif2008 done

internal/provider/onepassword_item_data_source.go
```diff
@@ -184,6 +194,56 @@ func (d *OnePasswordItemDataSource) Schema(ctx context.Context, req datasource.S
...
+										Sensitive:           true,
+									},
+									"content_base64": schema.StringAttribute{
+										MarkdownDescription: fileContentBase64Description,
+										Computed:            true,
+										Sensitive:           true,
+									},
+								},
+							},
+						},
```
edif2008, line 221: `[suggestion]`

Here's one last thing that would be worth doing, and apologies for not thinking about it sooner.

What would be helpful for maintainability purposes it to make this schema into a variable and then pass it to both the `section` and in the root parts of the schema. You can see the similar pattern being done with the `recipe` block. 

This would enable us to easily update this schema in both places by making only one change. This also ensures that the two blocks in the schema are consistent.

sdahlbac:
@edif2008 fixed

PR #172, opened 2024-05-17
Cleanup old provider code

edif2008:
## Sumary

This PR does the following:
- Remove the `onepassword` directory which contained the old implementation of the provider (with `terraform-plugin-sdk`)
- Remove the commented code related to old implementation in `main.go`
- Remove `github.com/hashicorp/terraform-plugin-sdk/v2` package from `go.mod` and vendor.
- Update packages to latest versions.

No testing is needed since this doesn't affect the current build of the provider.


jillianwilson (approved):
I have verified that the files removed are the ones that are expected to be moved after our migration to the terraform provider framework

Issue #174, opened 2024-05-21
v2.0.0 failing to apply with example code from the docs


bryanhonof:
### Your environment

Terraform Provider Version: `v2.0.0`

Connect Server Version: `n/a`

CLI Version: `2.24.0`

OS: `darwin_arm64`

Terraform Version: `v1.8.3`

## What happened?

```terraform
data "onepassword_vault" "vault" {
  name = "test"
}

resource "onepassword_item" "demo_login" {
  vault = data.onepassword_vault.vault.id

  title    = "Demo Terraform Login"
  category = "login"
  username = "test@example.com"
}
```

```console
$ terraform apply
...
onepassword_item.demo_login: Creating...
╷
│ Error: Plugin did not respond
│
│ The plugin encountered an error, and failed to respond to the plugin6.(*GRPCProvider).ApplyResourceChange call. The
│ plugin logs may contain more details.
╵

Stack trace from the terraform-provider-onepassword_v2.0.0 plugin:

panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x2 addr=0x60 pc=0x100499d00]

goroutine 81 [running]:
github.com/1Password/terraform-provider-onepassword/internal/provider.itemTerraformID(...)
	github.com/1Password/terraform-provider-onepassword/internal/provider/util.go:17
github.com/1Password/terraform-provider-onepassword/internal/provider.itemToData({0x1009fb058, 0x1400026ccf0}, 0x0, 0x140001aa4e0)
	github.com/1Password/terraform-provider-onepassword/internal/provider/onepassword_item_resource.go:458 +0x40
github.com/1Password/terraform-provider-onepassword/internal/provider.(*OnePasswordItemResource).Create(0x1400026ab50, {0x1009fb058, 0x1400026ccf0}, {{{{0x100a001e8, 0x140004672f0}, {0x1008e7240, 0x14000466750}}, {0x100a01d50, 0x140000615e0}}, {{{0x100a001e8, ...}, ...}, ...}, ...}, ...)
	github.com/1Password/terraform-provider-onepassword/internal/provider/onepassword_item_resource.go:342 +0x278
github.com/hashicorp/terraform-plugin-framework/internal/fwserver.(*Server).CreateResource(0x140001489c0, {0x1009fb058, 0x1400026ccf0}, 0x140003e55b8, 0x140003e5590)
[...]

PR #178, opened 2024-05-29
Throw error when 1Password client item creation fails

edif2008:
## Summary

We add to diagnostics the error returned when the 1Password client fails to create an item. However, we weren't stopping the process, which caused the provider to try to add the "failed" created item in the state which led to a crash of the provider.

Relates: #174

## How to test

1. Checkout this branch:
   ```sh
   git pull && git checkout fix/throw-err-on-item-creation-fail
   ```
2. Build the provider:
   ```sh
   make build
   ```
3. Ensure that [your environment is configured to use the local provider](https://developer.hashicorp.com/terraform/tutorials/providers-plugin-framework/providers-plugin-framework-provider#prepare-terraform-for-local-provider-install). Specifically, your `.terraform.rc` on macOS/Linux or `terraform.rc` on Windows contains this:
   ```
   provider_installation {
     dev_overrides {
         # other overrides
         "1Password/onepassword" = "path/to/terraform-provider-onepassword/dist"
     }
     # For all other providers, install them directly from their origin provider
     # registries as normal. If you omit this, Terraform will _only_ use
     # the dev_overrides block, and so no other providers will be available.
     direct {}
   }
   ```
4. Create a `main.tf` file with this content:
   ```tf
   terraform {
     required_providers {
       onepassword = {
         source = "1Password/onepassword"
         version = "~> 2.0"
       }
     }
   }

   data "onepassword_vault" "vault" {
     name = "demo"
   }

   resource "onepassword_item" "demo_login" {
     vault = data.onepassword_vault.vault.id

     title    = "Demo Terraform Login"
     category = "login"
     username = "test@example.com"
   }
   ```
5. Export the necessary environment variables for the authentication method that you want (Connect, service account or regular user)
6. Run `terraform apply`
   - [ ] The provider **should not** crash
   - [ ] The provider should throw an error like this one:
      ```
      ╷
      │ Error: 1Password Item create error
      │ 
      │   with onepassword_item.demo_login,
      │   on main.tf line 14, in resource "onepassword_item" "demo_login":
      │   14: resource "onepassword_item" "demo_login" {
      │ 
      │ Error creating 1Password item, got error op error: unable to process line 1: "vaults/ba7ntbqcum4tsrcc5ciem5wfsu" isn't a vault in this account. Specify the vault with its ID or name.
      ╵
      ```

### DIFF
--- a/internal/provider/onepassword_item_resource.go
+++ b/internal/provider/onepassword_item_resource.go
@@ -337,6 +337,7 @@ func (r *OnePasswordItemResource) Create(ctx context.Context, req resource.Creat
 	createdItem, err := r.client.CreateItem(ctx, item, item.Vault.ID)
 	if err != nil {
 		resp.Diagnostics.AddError("1Password Item create error", fmt.Sprintf("Error creating 1Password item, got error %s", err))
+		return
 	}
 
 	resp.Diagnostics.Append(itemToData(ctx, createdItem, &data)...)


back-2-95:
I tested this PR and provider did not crash anymore. But I'm having issues with inconsistent results (they vary on different results) AND defining vault (tried name, id, uuid).

edif2008:
Hey @back-2-95!

It sounds like this PR addressed the issue of the provider crashing, which is its purpose.
What you're describing seems to be something else that might be better suited for a separate issue. Would you be kind to open a new issue with the following details:
- The use case (can provide snippets that can be used to reproduce the issue you're facing)
- What you were expecting to get
- What is the actual outcome.

back-2-95:
Yes, I think those other issues are already mentioned in some other issues by others.

volodymyrZotov:
✅ Tested this and confirmed that the provider doesn't crash anymore.

volodymyrZotov (approved):

PR #180, opened 2024-05-30
Improve examples

edif2008:
This PR brings the following improvements in the example snippets:
- Make the data source snippets work on their own (no longer rely on a variable)
- Show only one resource example snippet
- Put the complex example with multiple item resources in a new directory
- Add the example of creating a Secure Note item
- Bump provider version in documentation to 2.0.0.

Resolves: #174


edif2008:
@bryanhonof How do these example snippets in this PR look to you? Are they clearer?

I've tested them myself and they work as expected.

bryanhonof (approved):
Didn't get to test it yet, but this already looks clearer then it was. 😁

jillianwilson (approved):

PR #196, opened 2024-08-23
Export provider initialization function

volodymyrZotov:
In order to be able to use `New` function from `internal` package in the pulumi-provider, it's exported within `providerlink` package now.


edif2008 (approved):
That is a great workaround. Let's hope this smoothens our process of using this version of the provider with the Pulumi bridge.

PR #199, opened 2024-10-11
Add `onepassword_item_share` Resource Type

joraff:
This PR adds item sharing as a managed resource!

Sharing an item is only supported by the 1Password CLI, and therefore, only by service and user accounts. Attempting to create this resource when using 1Password Connect Server will result in an error.


joraff:
Commits were signed, but I didn't have my signing key in GitHub - fixed!

P.S. - I realize this isn't that valuable of a resource type to have since it isn't _actually_ managing anything (read, update, and delete methods do nothing - by design), but I had a use for it. I debated for a hot minute between doing this and wrapping `op item share` in a `null_resource`, and decided to do this because I wanted the practice writing in go and plugin code! And it's prettier.

joraff:
Can a maintainer re-run the signed commits workflow? Should show as all signed now. Also, some feedback on the feature enhancement proposal?

joraff:
Going to open a new PR to clear the workflow failure

Issue #201, opened 2024-10-25
`onepassword_item` update behaviour  


odise:
### Your environment

Terraform Provider Version: 2.1.2

OS: macOS Sonoma

Terraform Version: 1.5.7

## What happened?

The current implementation of the provider does not handle manually item property changes very well. Mainly for `username` and `url` the behaviour is strange in various ways. Example:

```terraform
resource "onepassword_item" "test" {
  vault = data.onepassword_vault.acceptance-tests.uuid
  title = "TestItemResourceIntegrationUsername-D2AbjK"
  category = "login"
  username = null
  password_recipe {}
  url = null
}
```

1. Setting `username` and `url` to `null` will not touch the item property in 1Password.  Thats desired behaviour.
2. Setting `username` and `url` to a valid String e.g.`"xxx"` will overwrite underlaying values. Thats also expected.
3. Setting `username` and `url` to `""` will result in `When applying changes to onepassword_item.test, provider produced an unexpected new value: .username: was cty.StringVal(""), but now null` error message. See below.
4. Setting `username` and `url` to `null` after manually changed those properties beforehand will result in overwriting the value in 1Password. This is not expected (see below).

## What did you expect to happen?

The following behaviour would be expected based on the scenarios 3 and 4 from above:
* `username` and `url` = `""`: this should "reset" the property in 1Password i.e. the value of this property should be unset after apply.
* `username` and `url` = `null` or not set in resource: should leave the value for the item in 1Password untouched i.e. ignored by Terraform.

## Steps to reproduce

Imagine the following test scenario:

1. create an item leaving `username` and `url` unset in `onepassword_item`
2. manually set those item properties
3. reapply the resource -> no changes should be planed
4. set the item properties in Terraform e.g. `username = "test"` and reapply -> properties should be set by Terraform
5. unset the item properties in Terraform e.g. `username = ""` and reapply -> properties should be empty (unset i.e. without value) again

## Notes & Logs

We use the provider to create items that are populated with values derived directly from Terraform resources as well as placeholder items with predefined properties that are not set and needs to stay untouched by Terraform. The 1.4 version of the provider was able to handle those scenarios however it disappear with version 2. 

Sadly the unit tests implemented are not very flexible and it seems to me that some integration tests will be needed to make sure the behaviour is like described above. Do you have any integration tests available or planned? I'm happy to share mine ;).

JillRegan:
[...]

PR #202, opened 2024-10-25
item update behaviour to ignore manual changes

odise:
This PR addresses #201 for `username` and `url` item properties. Please let me know what you think.


JillRegan:
Hi @odise 👋 

I took a look at this PR and the issue. In regards to the null behaviour, making null ignore existing values would be a breaking change. Currently, null or omitting the attribute removes the field in 1Password.

But as for the empty string behaviour a fix for this was implemented and released in [v3.0.1](https://github.com/1Password/terraform-provider-onepassword/releases/tag/v3.0.1). So you are now able to set the fields username and url to "" in order to clear them! I am going to close the PR and issue but please feel free to tag me if you have any other questions or want to discuss this further!

PR #205, opened 2024-11-14
 Add `onepassword_item_share` Resource Type

joraff:
This PR adds item sharing as a managed resource!

Sharing an item is only supported by the 1Password CLI, and therefore, only by service and user accounts. Attempting to create this resource when using 1Password Connect Server will result in an error.


SimonBarendse:
Hi @joraff , thank you for the contribution! 🙌 

I'm thinking about how to fit item sharing into the idempotent design of Terraform, and how it will work with state drift. I believe the current use cases and interface design for item sharing are designed around sharing as a point in time action, and I anticipate challenges when treating it as a resource.

For example:
- **Expiry**: The relative definition "expires in" (e.g. 3 days), what is expected when applying again? Should a new link be generated that expires in 3 days from the new applied time? This feels in violation with Terraform's idempotent design.
    - If we'd have a resource based API I could imagine defining an "expires on" (e.g. April 14, 2025 0:00 UTC), which allows idempotent definitions. But even then, when the share is expired, how do we differentiate from a deleted share?
- **View Once**: This has a similar drift question. Viewing the share changes the state of the share (as it will expire/delete). How does this work with Terraform idempotancy?
- **Item Changes**: Item share links create a copy of the item at the time of sharing. When changes are made to the item (e.g. in the same Terraform project), should this propagate to a new link being generated with the updated contents? If you would be willing to share how you're using the item and item_share resources together I think that will be useful too.

I'm curious to learn more about your use case for using item sharing with Terraform. For example, when and how do the share links leave Terraform (e.g. is that also a point in time action or could/should it re-occur on changes)? And about your expectations and perspective on how to reconcile the point-in-time nature of item sharing with how it'll work in Terraform?

I'll also reach out here internally at 1Password to the team who owns and build the item sharing functionality to get their perspective too.

Thanks again for the contribution! 💙  And I'm looking forward to figuring out together if we can make this fit and if so how.

PR #158, opened 2024-03-04
feat: add ssh public key data source

atammy-narmi:
Relates: https://github.com/1Password/terraform-provider-onepassword/issues/74

PR #210, opened 2025-02-12
Fix credential item category and add another credential fields

speto:
I believe part of the problem with the credential category lies in [`if f.ID == "credential" && item.Category == "API_CREDENTIAL"`](https://github.com/speto/terraform-provider-onepassword/commit/b0ed23bb8b25781eea5688ab19bccbd9ff224d06#diff-18c55ffdfc96b7b20ba10dc130135cec8cb91111e4e4eb711994c18a71017533R366) where `"credential"` is checked against `f.ID`. 

I'm not sure if I'm following correctly everything, but I believe the correct `f` field for this comparison should be `Label` not `ID`. Leaving it like rest of the fields on [350-362](https://github.com/speto/terraform-provider-onepassword/commit/b0ed23bb8b25781eea5688ab19bccbd9ff224d06#diff-18c55ffdfc96b7b20ba10dc130135cec8cb91111e4e4eb711994c18a71017533R350) should be sufficient.

I followed discussion in the [original PR](https://github.com/1Password/terraform-provider-onepassword/pull/158) but I don't feel much clearer on it. Just think this shouldn't have been merged without tests.

So, first I added a test for credential category, which successfully failed. Then I tried fixing the issue and discovered the `f.ID` problem. Afterwards I added additional credential fields based on what 1password returns for that category and updated the docs. The `type` field is shared between both database and credential category. 

This is also related to https://github.com/1Password/terraform-provider-onepassword/issues/52


JillRegan:
Hi @speto 👋 I just tried to reproduce this issue and was able to access the credential field of an API Credential. It looks like this was resolved in [version 2.1.0](https://github.com/1Password/terraform-provider-onepassword/releases/tag/v2.1.0). 

I am going to close this PR but if you continue to encounter any more issues please tag me with some context and I can reopen!

speto:
Thanks @JillRegan.

While the credential field is accessible now, I believe the current implementation is incomplete and inconsistent with the codebase.

The condition `if f.ID == "credential" && item.Category == "API_CREDENTIAL"` breaks the parsing pattern of the immediate context. The rest of the logic handles fields via a `switch` statement. Injecting a specific `if` block for one category/field combination is technical debt that makes the parsing logic harder to read and maintain.

`API_CREDENTIAL` items are not just about the `credential` field. They specifically define `filename`, `valid_from`, and `expires` as well. My PR mapped these fields correctly. The current implementation misses them entirely.

Eventually, this should be addressed to align the parsing logic and include the missing fields.

JillRegan:
Thank you for your feedback, @speto. I appreciate you pointing out these issues.

I'm going to reopen this PR for review. We've added E2E tests since the PR was originally opened so we will want to look at adding testing there for these updates. 

We're planning to address some open issues in the Terraform Provider over the coming weeks, and I'll try to address these updates/concerns as part of that work. I can take a closer look at these changes once we tackle some existing bugs. 👍

internal/provider/const.go
```diff
@@ -17,15 +17,19 @@ const (
...
+	credentialDescription = "(Only applies to the API credential category) API credential for this item."
+	validFromDescription  = "(Only applies to the API credential category) The timestamp from which the API credential is valid."
+	expiresDescription    = "(Only applies to the API credential category) The timestamp when the API credential expires."
+	filenameDescription   = "(Only applies to the API credential category) The filename associated with the API credential."
+
 	dbHostnameDescription = "(Only applies to the database category) The address where the database can be found"
 	dbDatabaseDescription = "(Only applies to the database category) The name of the database."
 	dbPortDescription     = "(Only applies to the database category) The port the database is listening on."
-	dbTypeDescription     = "(Only applies to the database category) The type of database."
+	dbTypeDescription     = "(Only applies to database and API credential categories) For the database category"
```
JillRegan: Since this no longer would apply to just the database item we may want to update this to `typeDescription` instead of `dbTypeDescription`.  And possibly change to "(Only applies to database and API credential categories) The type of database or API Credential".

We will also want to update the `MarkdownDescription` for type in the schema around line ~160 to reflect that these changes mean type is no longer just database types.

internal/provider/onepassword_item_data_source.go
```diff
@@ -375,12 +390,16 @@ func (d *OnePasswordItemDataSource) Read(ctx context.Context, req datasource.Rea
 					data.PublicKey = types.StringValue(f.Value)
 				case "private key":
 					data.PrivateKey = types.StringValue(f.Value)
+				case "credential":
+					data.Credential = types.StringValue(f.Value)
+				case "valid_from":
```
JillRegan: I tested this using Connect and was not able to access the `valid_from` field. I believe this is because the label for this field is actually `valid from` and not `valid_from`.

internal/provider/onepassword_item_data_source.go
```diff
@@ -375,12 +390,16 @@ func (d *OnePasswordItemDataSource) Read(ctx context.Context, req datasource.Rea
...
+					data.ValidFrom = types.StringValue(f.Value)
+				case "expires":
+					data.Expires = types.StringValue(f.Value)
```
JillRegan: We may want to convert `expires` and `valid from` to date to match existing for resource items. So if field `f.Type` is `op.FieldTypeDate` then we use existing util `SecondsToYYYYMMDD` to convert to YYYY-MM-DD.

volodymyrZotov (approved):
✅ Great work! Approved!

PR #220, opened 2025-05-21
Remove vendor folder

volodymyrZotov:
This MR removes the vendor directory as it's not used in the CI/CD process and the project already uses go modules, so it's not necessary to keep `vendor`.

It also will make MR review way easier, after changing dependencies it won't make enormous changes in the `vendor` folder.


SimonBarendse (approved):
Jup, this should be fine to remove.

HashiCorp has also done this with the Terraform project itself and the providers they maintain themselves. See for example: https://github.com/hashicorp/terraform/pull/26358 and https://github.com/hashicorp/terraform-provider-aws/pull/15391

PR #229, opened 2025-09-28
Add write-only password to item resource

tbarthel-fr:
This commit introduces a write-only password feature to the `onepassword_item` resource. This allows for setting a password that is not stored in the Terraform state, which is useful for ephemeral values.

The following new attributes have been added:
- `password_wo`: A write-only password.
- `password_wo_version`: An integer that must be incremented to trigger an update to the `password_wo` field.

This commit also includes:
- Updates to various Go dependencies, including the Terraform plugin framework so that `schema.StringAttribute.WriteOnly` can be used.
- New acceptance tests for the write-only password functionality.


internal/provider/onepassword_item_resource.go
```diff
@@ -211,6 +213,32 @@ func (r *OnePasswordItemResource) Schema(ctx context.Context, req resource.Schem
 					ValueModifier(),
 				},
 			},
+			"password_wo": schema.StringAttribute{
+				MarkdownDescription: passwordWriteOnceDescription,
+				Optional:            true,
+				Sensitive:           true,
+				WriteOnly:           true,
```
volodymyrZotov, line 215: `WriteOnly` property not exists here.

jorhett:
> @tbarthel-fr Thank you for the contribution!
> I was able to review it and found that if using `password_wo` it still stored raw password in the plan.

Of course it is in the Plan. For a value to be used it must be in the Plan. That's how Terraform works.

The goal of this is to avoid it being stored in the State.

> WriteOnly property not exists here.

Sure it does. Here's the documentation https://developer.hashicorp.com/terraform/plugin/framework/resources/write-only-arguments#schema

Further, I'm a bit confused about why you make this statement. You generated a Plan, did you not? If you successfully generated the Plan, then that attribute is valid.

volodymyrZotov:
@jorhett The [build failed](https://github.com/1Password/terraform-provider-onepassword/actions/runs/19974940487/job/57288847209?pr=229).

Now I understand why. 

This PR doesn't bump `github.com/hashicorp/terraform-plugin-framework` to `v1.14.0`.

The current version `github.com/hashicorp/terraform-plugin-framework` used right now is  `v1.11.0` that doesn't have `WriteOnly` feature.

@tbarthel-fr Would you have a chance to update dependencies and pus the changes?

nikaro:
@volodymyrZotov @tbarthel-fr I made a PR on this branch: https://github.com/tbarthel-fr/terraform-provider-onepassword/pull/1

volodymyrZotov:
@nikaro thank you!

@tbarthel-fr could you merge it in please?

volodymyrZotov:
@nikaro As @tbarthel-fr didnt reply, I fixed it directly in this branch, thanks for the effort though.

volodymyrZotov (approved):
@tbarthel-fr I've polished it slightly. Thank you for the contribution ⭐️

PR #233, opened 2025-10-24
Add data source tests for ssh key and document items

JillRegan:
This PR adds E2E tests for the onepassword_item data source for SSH Key and Documents types with both lookup methods (Title and UUID).

### 🎯 Changes

- Update test cases in TestAccItemDataSource to use a helper in order to reduce repetitive test data.
- Add Document and SSHKey entries to testItems.
- Add test/e2e/utils/test_helpers.go with ValidateSSHKeys to parse/validate public and private SSH keys.

### 🔧 Benefits

- Less duplication; easier to add item types or lookup methods.
- Adds coverage for Document and SSH key items.

### DIFF
--- a/test/e2e/item_data_source_test.go
+++ b/test/e2e/item_data_source_test.go
@@ -10,7 +10,7 @@ import (
 	"github.com/hashicorp/terraform-plugin-testing/terraform"
 
 	tfconfig "github.com/1Password/terraform-provider-onepassword/v2/test/e2e/terraform/config"
-	"github.com/1Password/terraform-provider-onepassword/v2/test/e2e/utils"
+	"github.com/1Password/terraform-provider-onepassword/v2/test/e2e/utils/ssh"
 )
 
 const testVaultID = "bbucuyq2nn4fozygwttxwizpcy"
@@ -146,7 +146,7 @@ func TestAccItemDataSource(t *testing.T) {
 					publicKey := item.Primary.Attributes["public_key"]
 					privateKey := item.Primary.Attributes["private_key"]
 
-					return utils.ValidateSSHKeys(publicKey, privateKey)
+					return ssh.ValidateSSHKeys(publicKey, privateKey)
 				}))
 			}
 
--- a/test/e2e/utils/ssh/ssh.go
+++ b/test/e2e/utils/ssh/ssh.go
@@ -1,4 +1,4 @@
-package utils
+package ssh
 
 import (
 	"fmt"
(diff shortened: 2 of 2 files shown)


test/e2e/item_data_source_test.go
```diff
@@ -160,6 +135,21 @@ func TestAccItemDataSource(t *testing.T) {
 				checks = append(checks, resource.TestCheckResourceAttr("data.onepassword_item.test_item", attr, expectedValue))
 			}
 
+			// Validate SSH keys
```
JillRegan, line 138: Added separate logic using [this ssh package](https://pkg.go.dev/golang.org/x/crypto/ssh) to check if the public and private keys are valid instead of checking for exact match.

Did not want to add a hard coded ssh key to check exact match as felt that would be bad practice. Also wanted to ensure we weren't just checking if the attribute is set but confirming it is a valid key.

test/e2e/utils/test_helpers.go
```diff
@@ -0,0 +1,31 @@
+package utils
```
volodymyrZotov: What do you think about calling this package `ssh`? File can also be called `ssh.go` or `validate.go`.

That name clearly reflects what the package is for and can be extended with other ssh related function if needed.

test/e2e/item_data_source_test.go
```diff
@@ -160,6 +135,21 @@ func TestAccItemDataSource(t *testing.T) {
...
+			if tc.item.Attrs["category"] == "ssh_key" {
+				checks = append(checks, resource.TestCheckFunc(func(s *terraform.State) error {
+					item, ok := s.RootModule().Resources["data.onepassword_item.test_item"]
```
volodymyrZotov, line 141: `[question]`

Does this read an item from terraform state?

JillRegan: Yes. 👍 Because we are doing custom validation here we need to pull the item from terraform state. 

`TestCheckResourceAttr` from the terraform-plugin-testing package handles retrieving state the same way.

test/e2e/utils/test_helpers.go
JillRegan: My only thought here was if in the future we may end up using this package for other helpers. And in that case `ssh` or `validate `may not be relevant to those helpers. 

I think changing the file name to reflect that this is specific for `ssh` makes sense though, instead of calling it `test_helpers`. Other helpers can be separated out into their own packages in the future. I will make that update!

Issue #208, opened 2024-12-30
Feature Request: Support Password Recipe for Database Category using 1password cli


codezninja:
### Summary

Trying to create a database category in terraform one password item. But when using `password_receipe` the generated doesn't have any value. Here's a simple example I would expect the password field to generate a password

```
resource "onepassword_item" "test" {
  vault    = data.onepassword_vault.this.uuid
  category = "database"
  title    = "Test PSQL Login"

  password_recipe {
    length  = 64
    symbols = false
  }
}
```

### Use cases

Using category database since the item being created is to access a database

### Proposed solution
n/a

### Is there a workaround to accomplish this today?
1. Use random password generate in terraform to generate password
2. Save password using `password` attribute on item instead

### References & Prior Work
n/a

### Versions

```
Terraform v1.10.3
on darwin_arm64
+ provider registry.terraform.io/1password/onepassword v2.1.2
+ provider registry.terraform.io/cyrilgdn/postgresql v1.25.0
```

moward98:
Hi @codezninja, in order for us to better understand this issue, can you please share how you're attempting to create this database item? Is it via a Connect server, with a service account, or locally using biometrics?

codezninja:
> locally using biometrics

Hi @moward98 so I'm using op locally using biometrics

moward98:
Unfortunately this is currently the expected functionality since the CLI can **only** generate passwords for the password field within Login or Password items.

codezninja:
Ahh ic!! Is there a repo to make that feature request then for cli to add support for that?

codezninja:
> Please feel free to open an issue as a feature request here: [1Password/terraform-provider-onepassword/issues](https://github.com/1Password/terraform-provider-onepassword/issues?q=sort%3Aupdated-desc+is%3Aissue+is%3Aopen)

@moward98 since that is this same repo?

moward98:
@codezninja Yes that's fine

volodymyrZotov:
This is supported in the provider `v3.x`

PR #235, opened 2025-10-28
Add resource item tests

JillRegan:
This PR adds E2E tests for the onepassword_item resource item for Login, Password, Database, and Secure Note types.

### 🎯 Changes

- Add TestAccItemResource covering create, read, update, and delete for Login, Password, Database, and Secure Note categories. (Document and SSHKey to come in future PR)
- Add logStep helper to log readable lifecycle markers (CREATE/READ/UPDATE/DELETE) to test output, making test runs easier to follow and debug.
- Add test data maps (testItemsToCreate, testItemsUpdatedAttrs) and helpers to validate attributes and lifecycle steps.
- Add ItemResourceConfig builder (item_resource_config.go) to generate resource configs (supports optional password_recipe).
- Rename CreateItemDataSourceConfigBuilder -> CreateConfigBuilder and update usages.
- Update existing tests to use the new CreateConfigBuilder.

### 🔧 Benefits

- Adds full CRUD E2E coverage for onepassword_item across multiple categories.

### DIFF
--- a/test/e2e/item_resource_test.go
+++ b/test/e2e/item_resource_test.go
@@ -110,8 +110,23 @@ func TestAccItemResource(t *testing.T) {
 	for _, tc := range testCases {
 		t.Run(tc.name, func(t *testing.T) {
 			item := testItemsToCreate[tc.category]
-
 			var itemUUID string
+
+			// Build check functions for create step
+			createChecks := []resource.TestCheckFunc{
+				logStep(t, "CREATE"),
+				captureItemUUID(t, "onepassword_item.test_item", &itemUUID),
+			}
+			bcCreate := buildItemChecks("onepassword_item.test_item", item.Attrs)
+			createChecks = append(createChecks, bcCreate...)
+
+			// Build checks for update step
+			updateChecks := []resource.TestCheckFunc{
+				logStep(t, "UPDATE"),
+				verifyItemUUIDUnchanged(t, "onepassword_item.test_item", &itemUUID),
+			}
+			bcUpdate := buildItemChecks("onepassword_item.test_item", testItemsUpdatedAttrs[tc.category])
+			updateChecks = append(updateChecks, bcUpdate...)
 
 			resource.Test(t, resource.TestCase{
 				ProtoV6ProviderFactories: testAccProtoV6ProviderFactories,
@@ -122,10 +137,7 @@ func TestAccItemResource(t *testing.T) {
 							tfconfig.ProviderConfig(),
 							tfconfig.ItemResourceConfig(testVaultID, item.Attrs),
 						),
-						Check: resource.ComposeAggregateTestCheckFunc(append([]resource.TestCheckFunc{
-							logStep(t, "CREATE"),
-							captureItemUUID(t, "onepassword_item.test_item", &itemUUID),
-						}, buildItemChecks("onepassword_item.test_item", item.Attrs)...)...),
+						Check: resource.ComposeAggregateTestCheckFunc(createChecks...),
 					},
 					// Read/Import new item and verify it matches state
 					{
@@ -144,10 +156,7 @@ func TestAccItemResource(t *testing.T) {
 							tfconfig.ProviderConfig(),
 							tfconfig.ItemResourceConfig(testVaultID, testItemsUpdatedAttrs[tc.category]),
 						),
-						Check: resource.ComposeAggregateTestCheckFunc(append([]resource.TestCheckFunc{
-							logStep(t, "UPDATE"),
-							verifyItemUUIDUnchanged(t, "onepassword_item.test_item", &itemUUID),
-						}, buildItemChecks("onepassword_item.test_item", testItemsUpdatedAttrs[tc.category])...)...),
+						Check: resource.ComposeAggregateTestCheckFunc(updateChecks...),
 					},
 					// Delete new item
 					{


test/e2e/terraform/config/config.go
```diff
@@ -21,3 +21,16 @@ func CreateItemDataSourceConfigBuilder() func(functions ...func() string) string
 		return configStr
 	}
 }
+
+func CreateItemResourceConfigBuilder() func(functions ...func() string) string {
```
volodymyrZotov: I think this is similar as `CreateItemDataSourceConfigBuilder`, so we can just rename the function to be more broad aka `CreateConfigBuilder`

test/e2e/item_resource_test.go
```diff
@@ -0,0 +1,198 @@
...
+		{op.SecureNote, "SecureNote"},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			item := testItemsToCreate[tc.category]
+
+			// Determine if password_recipe is supported for this category
+			// Only Login and Password support password_recipe currently
+			usePasswordRecipe := tc.category == op.Login || tc.category == op.Password
```
JillRegan: Currently these are the only categories that allows password recipes. There is an [open issue](https://github.com/1Password/terraform-provider-onepassword/issues/208) to add support for database.

test/e2e/item_resource_test.go
```diff
@@ -0,0 +1,198 @@
...
+					{
+						Config: tfconfig.CreateConfigBuilder()(
+							tfconfig.ProviderConfig(),
+							tfconfig.ItemResourceConfig(testVaultID, initialConfig, usePasswordRecipe),
+						),
+						Check: resource.ComposeAggregateTestCheckFunc(append([]resource.TestCheckFunc{
+							logStep(t, "CREATE"),
+						}, buildItemChecks("onepassword_item.test_item", initialConfig)...)...),
+					},
+					// Read/Import new item and verify it matches state
```
JillRegan, line 142: Use import to read the item as per documentation [here](https://registry.terraform.io/providers/1Password/onepassword/latest/docs/resources/item#import). Here we:

1. Call to use terraform import
2. Execute `terraform import onepassword_item.myitem vaults/<vault uuid>/items/<item name>`
3. Call to verify that the item matches with previous state (ignoring the `password_recipe` field)

Documentation for Import variables [here](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-sdk/helper/resource#TestStep) if needed.

test/e2e/item_resource_test.go
```diff
@@ -0,0 +1,202 @@
...
+						ExpectError: regexp.MustCompile("Unable to read item"),
+						Check:       logStep(t, "DELETE"),
+					},
+				},
+			})
+		})
+	}
+}
+
+// logStep logs the current test step for easier test debugging
```
JillRegan, line 181: Added logic to log each step for easier debugging. If a step fails we can see previous ones that passed.

== RUN   TestAccItemResource/Login
    item_resource_test.go:169: CREATE
    item_resource_test.go:133: READ
    item_resource_test.go:169: UPDATE
    item_resource_test.go:169: DELETE
=== RUN   TestAccItemResource/Password
    item_resource_test.go:169: CREATE
    item_resource_test.go:133: READ
    item_resource_test.go:169: UPDATE
    item_resource_test.go:169: DELETE
=== RUN   TestAccItemResource/Database
    item_resource_test.go:169: CREATE
    item_resource_test.go:133: READ
    item_resource_test.go:169: UPDATE
    item_resource_test.go:169: DELETE
=== RUN   TestAccItemResource/SecureNote
    item_resource_test.go:169: CREATE
    item_resource_test.go:133: READ
    item_resource_test.go:169: UPDATE
    item_resource_test.go:169: DELETE
--- PASS: TestAccItemResource (41.15s)
    --- PASS: TestAccItemResource/Login (9.84s)
    --- PASS: TestAccItemResource/Password (10.81s)
    --- PASS: TestAccItemResource/Database (10.16s)
    --- PASS: TestAccItemResource/SecureNote (10.35s)

test/e2e/item_resource_test.go
```diff
@@ -0,0 +1,202 @@
...
+		"tags":     "updatedTag",
+	},
+	op.SecureNote: {
+		"title":      "Test Secure Note Create - Updated",
+		"note_value": "This is an updated secure note",
+		"tags":       "updatedTag",
+	},
+}
+
+func TestAccItemResource(t *testing.T) {
```
JillRegan, line 99: Created this as 1 test with separate Create, Read, Update, Delete steps. If we ran as separate tests resource.Test() destroys created items during teardown, this would mean we would need to recreate the items during every tests which would increase test run time significantly.

test/e2e/terraform/config/item_resource_config.go
```diff
@@ -0,0 +1,36 @@
...
+}
+
+func ItemResourceConfig(vaultID string, params map[string]string, passwordRecipe bool) func() string {
+	return func() string {
+		resourceStr := `resource "onepassword_item" "test_item" {`
+
+		resourceStr += fmt.Sprintf("\n  vault = %q", vaultID)
+
+		for key, value := range params {
+			if key == "tags" {
```
JillRegan: Will be refactored in future PR for more specific testing on Tags. Added for basic testing in these changes.

test/e2e/terraform/config/item_resource_config.go
```diff
@@ -0,0 +1,36 @@
...
+				resourceStr += fmt.Sprintf("\n  %s = [%q]", key, value)
+				continue
+			}
+			resourceStr += fmt.Sprintf("\n  %s = %q", key, value)
+		}
+
+		if passwordRecipe {
```
JillRegan: Same as above. Added for basic password creation testing, will be modifying for more specific testing on password generation.

test/e2e/terraform/config/item_resource_config.go
```diff
@@ -0,0 +1,36 @@
+package terraform
+
+import "fmt"
+
+type ItemResource struct {
+	Params map[string]string
+}
+
+func ItemResourceConfig(vaultID string, params map[string]string, passwordRecipe bool) func() string {
```
volodymyrZotov: This function is fine for now, but I can see that we could refactor it. Instead of passing `params map[string]string` we can pass `params map[string]interface{}` (similar to `params map[string]any`), and then use `reflection` package to identify what is the type of value and process it accordingly.

test/e2e/terraform/config/item_resource_config.go
```diff
@@ -0,0 +1,36 @@
...
+			if key == "tags" {
+				resourceStr += fmt.Sprintf("\n  %s = [%q]", key, value)
+				continue
+			}
+			resourceStr += fmt.Sprintf("\n  %s = %q", key, value)
+		}
+
+		if passwordRecipe {
+			resourceStr += `
+			password_recipe {
```
volodymyrZotov: Can we pass `password_recipe` from outside as a map or other type and build it here?

test/e2e/item_resource_test.go
```diff
@@ -0,0 +1,202 @@
...
+	testCases := []struct {
+		category op.ItemCategory
+		name     string
+	}{
+		{op.Login, "Login"},
```
volodymyrZotov: JFYI: in Go, this implementation is fine as long as you keep the order parameters the same. But if you would swap it `{"Login", op.Login},`, it won't compile. But you can still fix it by setting property names `{name: "Login", category: op.Login},`

test/e2e/item_resource_test.go
```diff
@@ -0,0 +1,202 @@
...
+				Steps: []resource.TestStep{
+					// Create new item
+					{
+						Config: tfconfig.CreateConfigBuilder()(
+							tfconfig.ProviderConfig(),
+							tfconfig.ItemResourceConfig(testVaultID, initialConfig, usePasswordRecipe),
+						),
+						Check: resource.ComposeAggregateTestCheckFunc(append([]resource.TestCheckFunc{
+							logStep(t, "CREATE"),
+						}, buildItemChecks("onepassword_item.test_item", initialConfig)...)...),
```
volodymyrZotov: Could we extract standalone functions from `buildItemChecks` and apply which we need here?
So  it looks something like (some of functions may not exists in your implementation, just an example):
```
checkItemTags(),
checkItemPassword(),
checkCategory,
.... etc
```

test/e2e/item_resource_test.go
```diff
@@ -0,0 +1,225 @@
...
+					},
+					// Update new item
+					{
+						Config: tfconfig.CreateConfigBuilder()(
+							tfconfig.ProviderConfig(),
+							tfconfig.ItemResourceConfig(testVaultID, testItemsUpdatedAttrs[tc.category]),
+						),
+						Check: resource.ComposeAggregateTestCheckFunc(append([]resource.TestCheckFunc{
+							logStep(t, "UPDATE"),
+							verifyItemUUIDUnchanged(t, "onepassword_item.test_item", &itemUUID),
```
JillRegan: To ensure we did not just create a new item I have added a check to verify the the UUID matches between the item in the CREATE step and the item in the UPDATE step.

test/e2e/item_resource_test.go
```diff
@@ -0,0 +1,225 @@
...
+
+// logStep logs the current test step for easier test debugging
+func logStep(t *testing.T, step string) resource.TestCheckFunc {
+	return func(s *terraform.State) error {
+		t.Log(step)
+		return nil
+	}
+}
+
+// buildItemChecks creates a list of test assertions to verify item attributes
```
JillRegan, line 189: The below functions are moved to helper packages in the password generator [PR](https://github.com/1Password/terraform-provider-onepassword/pull/236/files).

test/e2e/terraform/config/item_resource_config.go
JillRegan: Updated to be able to handle any value in the map. When I updated to `interface{}` I received a warning to use `any ` instead. Read [here](https://medium.com/@csaba.ujvari/go-1-18-any-instead-of-interface-ce15dd7bea5d) it was added in Go 1.18.

test/e2e/terraform/config/item_resource_config.go
JillRegan: I ended up removing this check for now and handle the password recipe in this [pull request](https://github.com/1Password/terraform-provider-onepassword/pull/236) , as we add separate tests for this. Felt it was better handled there.

test/e2e/item_resource_test.go
JillRegan: I ended up updating to `{name: "Login", category: op.Login}`. I think given the similarity of these two parameters it is best to set the property names as I could see this may be confusing for anyone else looks to expand test cased.

test/e2e/item_resource_test.go
JillRegan: My only concern here would be that `buildItemChecks` loops through each data items provided attributes and builds the checks off of that. 

Example:

checkItemTags(),
checkItemPassword(),
checkCategory(),
checkUsername() -> Does not apply to the password item
checkDatabase() -> Only applies to database

With some of the other suggested changes I have updated the `buildItemChecks` to handle for string values and slices (for `tags`). But LMK your thoughts.

volodymyrZotov: It's so hard to read it now with several spread (`...`) operators.

Could this be refactored a bit to make it clearer what's going on, please?
```
createChecks := []resource.TestCheckFunc{
				logStep(t, "CREATE"),
				captureItemUUID(t, "onepassword_item.test_item", &itemUUID),
			}
bcCreate := buildItemChecks("onepassword_item.test_item", item.Attrs)
createChecks = append(createChecks, bcCreate...)


updateChecks := []resource.TestCheckFunc{
				logStep(t, "UPDATE"),
				verifyItemUUIDUnchanged(t, "onepassword_item.test_item", &itemUUID),
}
bcUpdate := buildItemChecks("onepassword_item.test_item", testItemsUpdatedAttrs[tc.category])
updateChecks = append(updateChecks, bcUpdate...)

....
Check: resource.ComposeAggregateTestCheckFunc(createChecks)

....


Check: resource.ComposeAggregateTestCheckFunc(updateChecks)
```

PR #236, opened 2025-10-30
Add password generation test

JillRegan:
This PR adds a  E2E test for password generation and refactors E2E test helpers to support password_recipe usage.

### 🎯 Changes

- Add `TestAccItemResourcePasswordGeneration` to validate password recipes (length, symbols, digits, letters).
- Add BuildPasswordRecipeChecks — validate password_recipe and generated password patterns.
- Support nested map -> block formatting in item_resource_config (for password_recipe).
- Move helpers into packages for refactoring

### 🔧 Benefits

- Adds targeted coverage for password generation logic and ensures generated passwords match requested recipes.
- Centralizes and reuses attribute and UUID assertions for future E2E tests

### DIFF
--- a/test/e2e/item_resource_test.go
+++ b/test/e2e/item_resource_test.go
@@ -225,7 +225,7 @@ func TestAccItemResourcePasswordGeneration(t *testing.T) {
 				if tc.recipe.Length < 1 || tc.recipe.Length > 64 {
 					testStep.ExpectError = regexp.MustCompile(`length value must be between 1 and 64`)
 				} else {
-					checks := password.BuildPasswordRecipeChecks("onepassword_item.test_item", recipeMap)
+					checks := password.BuildPasswordRecipeChecks("onepassword_item.test_item", tc.recipe)
 					testStep.Check = resource.ComposeAggregateTestCheckFunc(checks...)
 				}
 
--- a/test/e2e/utils/password/password_generation.go
+++ b/test/e2e/utils/password/password_generation.go
@@ -16,26 +16,21 @@ type PasswordRecipe struct {
 }
 
 // BuildPasswordRecipeChecks creates a list of test assertions to verify password recipe attributes
-func BuildPasswordRecipeChecks(resourceName string, recipe map[string]any) []resource.TestCheckFunc {
+func BuildPasswordRecipeChecks(resourceName string, recipe PasswordRecipe) []resource.TestCheckFunc {
 	checks := []resource.TestCheckFunc{
 		resource.TestCheckResourceAttr(resourceName, "password_recipe.#", "1"),
 	}
 
-	length, ok := recipe["length"].(int)
-	// If length is not provided, the default is 32
-	if !ok {
-		length = 32
+	length, symbols, digits, letters := recipe.Length, recipe.Symbols, recipe.Digits, recipe.Letters
+
+	// If all are false, default all to true (invalid recipe)
+	if !symbols && !digits && !letters {
+		symbols, digits, letters = true, true, true
 	}
 
-	symbols, _ := recipe["symbols"].(bool)
-	digits, _ := recipe["digits"].(bool)
-	letters, _ := recipe["letters"].(bool)
-
-	// If all attributes are false the provided recipe is invalid, so we default to true
-	if !symbols && !digits && !letters {
-		symbols = true
-		digits = true
-		letters = true
+	// If length is not provided (0), the default is 32
+	if recipe.Length == 0 {
+		length = 32
 	}
 
 	if length > 0 {


test/e2e/terraform/config/item_resource_config.go
```diff
@@ -32,6 +32,17 @@ func formatTerraformAttribute(key string, value any) string {
 		}
 		return fmt.Sprintf("\n  %s = [%s]", key, strings.Join(quotedItems, ", "))
 
+	case reflect.Map:
+		blockStr := fmt.Sprintf("\n  %s {", key)
+		attributes := value.(map[string]any)
```
volodymyrZotov: It will panic if provider key has a type other than `string`. We need to handle it:
```
attributes, ok := value.(map[string]any)
if !ok {
   // handle casting failure 
}
```

test/e2e/utils/checks/password_generation.go
```diff
@@ -0,0 +1,63 @@
+package checks
```
volodymyrZotov: `[nit]`

I'd call this file just `password.go` so any other logic related to password checks can be added here. Though, not sure if we ever will need it. Feel free to rename the file or leave as it is.

test/e2e/item_resource_test.go
```diff
@@ -178,57 +180,59 @@ func TestAccItemResource(t *testing.T) {
...
-		resource.TestCheckResourceAttrSet(resourceName, "id"),
-	}
+	// Test both Login and Password items
+	items := []op.ItemCategory{op.Login, op.Password}
 
-	for attr, expectedValue := range attrs {
-		// Check if the value is a slice and iterate over it
-		if slice, ok := expectedValue.([]string); ok {
-			checks = append(checks, resource.TestCheckResourceAttr(resourceName, fmt.Sprintf("%s.#", attr), fmt.Sprintf("%d", len(slice))))
+	for _, item := range items {
```
volodymyrZotov: `[nit]`

Most common way to write tests that I saw was, that you iterate through test cases and spin up test's specific cases. Another words iterate through the test cases and then nested loop iterates through categories.
```
for _, tc := range testCases {
   for _, item := range items {
      // tests logic
   }
}
```

With this order it seems easier to follow what's going on, as you first read what are the `testCases` and then you see that you iterate through tests right away
```
for _, tc := range testCases {
}
```

test/e2e/item_resource_test.go
```diff
@@ -178,57 +180,59 @@ func TestAccItemResource(t *testing.T) {
...
-func logStep(t *testing.T, step string) resource.TestCheckFunc {
-	return func(s *terraform.State) error {
-		t.Log(step)
-		return nil
+func TestAccItemResourcePasswordGeneration(t *testing.T) {
+	testCases := []struct {
+		name   string
+		recipe map[string]any
+	}{
+		{name: "Length32", recipe: map[string]any{"length": 32, "symbols": false, "digits": false, "letters": true}},
```
volodymyrZotov: - Is there a way to set `"length": 0`? What would happen in this case? Do we handle this?
- What if user provides empty map `map[string]any{}`? All boolean values will be set to `false` by default and integer will be set to `0` by default. Will it throw an error or generate a password with default length? In case of an error do we catch it?

test/e2e/item_resource_test.go
```diff
@@ -178,57 +180,59 @@ func TestAccItemResource(t *testing.T) {
...
 
-// logStep logs the current test step for easier test debugging
-func logStep(t *testing.T, step string) resource.TestCheckFunc {
-	return func(s *terraform.State) error {
-		t.Log(step)
-		return nil
+func TestAccItemResourcePasswordGeneration(t *testing.T) {
+	testCases := []struct {
+		name   string
+		recipe map[string]any
```
volodymyrZotov: Let's use a struct here instead of map, so the writer of the tests do not guess what properties are available.

test/e2e/utils/uuid/uuid.go
```diff
@@ -0,0 +1,51 @@
...
+import (
+	"fmt"
+	"testing"
+
+	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
+	"github.com/hashicorp/terraform-plugin-testing/terraform"
+)
+
+// CaptureItemUUID captures the UUID of a resource item
+func CaptureItemUUID(t *testing.T, resourceName string, uuidPtr *string) resource.TestCheckFunc {
```
volodymyrZotov: Why do we need to pass `uuidPtr` as a pointer? Why not to pass it as a string?

test/e2e/terraform/config/item_resource_config.go
JillRegan: In the case of a casting failure I think we should throw an error as they key should always be of type string. I updated the function to throw an error if the map type assertion fails, with a descriptive message. 

LMK if you feel we should handle this differently!

test/e2e/utils/checks/password_generation.go
JillRegan: Updated ✅

test/e2e/item_resource_test.go
JillRegan: Updated ✅

test/e2e/item_resource_test.go
JillRegan: I added a few test cases for the scenarios you mentioned. 

1. Length set as 0 - will throw and error as 0 is an invalid password length
2. Empty map - This defaults all character values to true, and length to default of 32
3. Just Length provided - Character values all default to true
4. All Character values set to false - this is invalid as at least 1 needs to be set to true. The false values are ignored and all revert to true.

test/e2e/item_resource_test.go
JillRegan: Updated here to use a struct. I had to use `*bool` and `*int` instead of `bool` and 'int` to distinguish between a value not being set and a value being explicitly false in order to test not passing length or a character attribute. 

```
type PasswordRecipe struct {
	Length  *int
	Symbols *bool
	Digits  *bool
	Letters *bool
}
```

test/e2e/terraform/config/item_resource_config.go
volodymyrZotov: I think this is fine! Thanks.

test/e2e/item_resource_test.go
volodymyrZotov: In Go, if you do not explicitly set property the values will be set by defaults:
- to `false` for boolean type
- to `0` for integer
- to `""` for string

You do not need to distinguish between values set to `false` and not set at all, as in both cases the property will have a values `false`. And therefore you do not need to use pointers in the struct types. This will simplify implementation a lot.

Example:
```
type PasswordRecipe struct {
	Length  int
	Symbols bool
	Digits  bool
	Letters bool
}

func main() {
	fmt.Printf("%+v", &PasswordRecipe{}) // will print "&{Length:0 Symbols:false Digits:false Letters:false}"
}
```

JillRegan: Updated ✅

test/e2e/item_resource_test.go
JillRegan: After discussion on slack, updated to just add test cases for length of 0, length of 65, and all character values set to false (as they end up reverting to true).

PR #238, opened 2025-11-05
Add sections and fields tests

JillRegan:
Add an E2E test for item sections and fields and update test helpers to support nested blocks and recursive attribute checks.

### 🎯Changes

- Add `TestAccItemResourceSectionsAndFields` to `test/e2e/item_resource_test.go` (create + update cases, UUID verification)
- Format nested map slices as Terraform blocks in `test/e2e/terraform/config/item_resource_config.go`
- Add recursive attribute checks in `test/e2e/utils/checks/attributes.go` to validate nested block lists

### 🔧Benefits

Enables E2E testing of sections → fields nesting
Makes test config generation and assertions handle nested blocks reliably

### DIFF
--- a/test/e2e/terraform/config/item_resource_config.go
+++ b/test/e2e/terraform/config/item_resource_config.go
@@ -31,7 +31,7 @@ func formatTerraformAttribute(key string, value any) (string, error) {
 	switch rv.Kind() {
 	case reflect.Slice, reflect.Array:
 		// Handle slices of maps recursively
-		if rv.Len() > 0 && rv.Index(0).Kind() == reflect.Map {
+		if rv.Type().Elem().Kind() == reflect.Map {
 			blockStr := ""
 
 			for i := 0; i < rv.Len(); i++ {


test/e2e/terraform/config/item_resource_config.go
```diff
@@ -26,6 +26,27 @@ func formatTerraformAttribute(key string, value any) string {
 
 	switch rv.Kind() {
 	case reflect.Slice, reflect.Array:
+		// Handle slices of maps recursively
+		if rv.Len() > 0 && rv.Index(0).Kind() == reflect.Map {
```
volodymyrZotov: This can be simplified:
```suggestion
        if rv.Type().Elem().Kind() == reflect.Map {
```

test/e2e/terraform/config/item_resource_config.go
```diff
@@ -26,6 +26,27 @@ func formatTerraformAttribute(key string, value any) string {
...
+			blockStr := ""
+
+			for i := 0; i < rv.Len(); i++ {
+				blockStr += fmt.Sprintf("\n  %s {", key)
+				attributes, ok := rv.Index(i).Interface().(map[string]any)
+
+				if !ok {
+					panic(fmt.Sprintf("invalid terraform config: attribute %q has unsupported type %T", key, value))
```
volodymyrZotov: We should return an error here and handle it, instead of panicking.

test/e2e/utils/checks/attributes.go
```diff
@@ -14,16 +14,36 @@ func BuildItemChecks(resourceName string, attrs map[string]any) []resource.TestC
...
 
-			for i, val := range slice {
-				checks = append(checks, resource.TestCheckResourceAttr(resourceName, fmt.Sprintf("%s.%d", attr, i), val))
+func buildAttributeChecks(resourceName string, attrPath string, expectedValue any) []resource.TestCheckFunc {
+	var checks []resource.TestCheckFunc
+
+	switch v := expectedValue.(type) {
+	case []string:
+		// Handle string slices
+		checks = append(checks, resource.TestCheckResourceAttr(resourceName, fmt.Sprintf("%s.#", attrPath), fmt.Sprintf("%d", len(v))))
```
volodymyrZotov: Seems you're trying to convert `len(v)` integer to string. For that you can use specific function designed for such conversion `strconv.Itoa(len(v))`

test/e2e/terraform/config/item_resource_config.go
JillRegan: Updated ✅

test/e2e/terraform/config/item_resource_config.go
JillRegan: Updated ✅ Also addressed other panic in the function.

test/e2e/utils/checks/attributes.go
JillRegan: Thank you, updated 👍

volodymyrZotov (approved):

PR #240, opened 2025-11-05
Add tags tests

volodymyrZotov:
This PR adds tests for item `tags`.

1. It creates item with 2 tags
2. Adds another one tag
3. Removes 2 tags

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,3 +33,5 @@ override.tf.json
 
 **/terraform-provider-onepassword
 /dist
+.idea
+.DS_Store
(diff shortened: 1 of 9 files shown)


JillRegan:
These look great 👍 I am wondering if we should make these table driven to align with our other tests, as well as make it easier for future expansion?

We could remove the tags attribute from each item in `testItemsToCreate` and just append them on via test cases. Something like: 
```
testCases := []struct {
name string
tags []string
}{
  {"CREATE_ITEM_WITH_2_TAGS", []string{"firstTestTag", "secondTestTag"}},
  {"ADD_3RD_TAG", []string{"firstTestTag", "secondTestTag", "thirdTestTag"}},
  {"REMOVE_2_TAGS", []string{"firstTestTag"}},
}

// rest of logic

// In a for loop
attrs["tags"] = step.tags
checks := []resource.TestCheckFunc{logStep(t, step.name)}
checks = append(checks, checks.BuildItemChecks("onepassword_item.test_item", attrs)
```

PR #244, opened 2025-11-06
Fix eventual consistency issue in Connect's item get implementation

volodymyrZotov:
The current implementation of Connect's `GetItem` method has an issue as it could not find an item which has been just created via Connect. As connect has some delay between item creation and when it becomes available for reading. and terraform make subsequent read right after create.

Therefore implementation is changed to check if provided UUID is a valid UUID. If so, it tries to get item by UUID with retries. If could not found after 5 retries, it will tru to get item by title.

This fix will enable running e2e tests using Connect as auth method as well.

### DIFF
--- a/internal/onepassword/connect/connect_client.go
+++ b/internal/onepassword/connect/connect_client.go
@@ -2,9 +2,13 @@ package connect
 
 import (
 	"context"
+	"strings"
+	"time"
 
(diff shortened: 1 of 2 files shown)


JillRegan (approved):
Thank you! ⭐

Issue #222, opened 2025-06-10
(409) Conflict: Internal server conflict.


leetrout:
### Your environment

Terraform Provider Version: 

`2.1.2`

<details>
<pre>
<code>
provider "registry.terraform.io/1password/onepassword" {
  version     = "2.1.2"
  constraints = "~> 2.1.2"
  hashes = [
    "h1:GdQ0kRDSi6naVt37Ui42uGLpgDsIEGx8Q8qBC7Ja2vo=",
    "zh:03d20138bf7bc645707b2c0c00203f66c07902d03c72be3f5f7bc365155bdc35",
    "zh:0bf54b246f141a7d0cb75c7c2c086d372c810efc061bf5a7ae0b62b70d9558f4",
    "zh:0ee19a8d1c193eaacc9679fb5ccf1d2be5d0c5e4173c3f3d82c09c717d3f354e",
    "zh:152c35cdd1bb98683c0a24e48b286eab0473735da242aa27b39df81bc6f84b63",
    "zh:293d264bccae325cdadcc9125b1d9fe9947a06cab05abb27c301d6244ef24cfd",
    "zh:37ade0a6cc2c8c2a15535f83281caf54f6fcbc9c75f0588407f1dd8da04a9a95",
    "zh:3c28c2352cd12464543e95b7d0b827abfe42794e9779f4e049789d864f88f3af",
    "zh:4676825834b3132234250046ad888b881ad54f6369b5f0302c6b3250ad4983b1",
    "zh:7da0327fe81c50bb71b510b342a5e2b90a3129f0e30dc479947b50e1dba9fa0d",
    "zh:890df766e9b839623b1f0437355032a3c006226a6c200cd911e15ee1a9014e9f",
    "zh:bb2899cada1cb964d75a9a4a7b915ac1ea03131769933e04b20ed0793f8eb02b",
    "zh:d142810d6f04973abf8471a2cc2fb09537f79010a0f20bd62866068a42fa5ab7",
    "zh:e2a39b685acd7b6846c8da3a94a355964ad2b411dc68c058a31dd2ce56aadd4a",
    "zh:effe9dea248ae4fb540ca5c84e2e369b467e229dc1518c606d3aa714049daa74",
    "zh:fbc4b2d7f7055a1b35c573e93bf89c6cad5f47f5dc30c0e2c61d5abfcf083cca",
  ]
}
</code>
</pre>
</details>

Connect Server Version: 

[...]

PR #246, opened 2025-11-11
Eventual consistency for Connect client

volodymyrZotov:
### ✨ Summary
This PR improves Connect client implementation by adding a delay after successful create, update or delete.

This is done because Connect have a delay when modifying item in 1Password server and local database, which is managed by Connect Sync service. Therefore subsequent Read will return old item. Which leads to the refresh plan issues. Therefore we need to wait before local db will have most up to date item. 

To address this new `wait` function is introduced. It gets an item from local database and makes sure it was updated, therefore we can proceed with updating terraform state. `wait` uses exponential backoff to increase wait time after each GetItem request re-retry.

The delay can be configured by setting `MaxWaitTime` and `MaxRetries` values in Connect config.

This PR also fixes `ProviderUserAgent` value for Connect. Previously it was set to `OpCLIPath`, which is not correct, and now we set it to `fmt.Sprintf("terraform-provider-onepassword/%s", p.version)`.

### 🔗 Resolves:
https://github.com/1Password/terraform-provider-onepassword/issues/222

### ✅ Checklist

- [x] 🖊️ Commits are signed
  - [ ] 🔹 Unit /🔸 Integration
  - [ ] 🌐 E2E

All three operations (Create, Update, Delete) follow the pattern:
- Send changes to the 1Password service
- The Sync service maintains the local SQLite database by syncing changes from the remote service.

That leads to eventual consistency issues.

### DIFF
--- a/internal/onepassword/connect/connect_client.go
+++ b/internal/onepassword/connect/connect_client.go
@@ -13,13 +13,6 @@ import (
 )
 
 type Config struct {
-	// MaxWaitTime is the maximum total time to wait for Connect to propagate changes
-	// to the local SQLite database after Create, Update, or Delete operations.
-	// The wait function uses exponential backoff with up to MaxRetries attempts to verify
-	// that changes have propagated. If not set, defaults to 5 seconds.
-	// This ensures eventual consistency - the sync service needs time to sync changes
-	// from the remote service to the local database that the API reads from.
-	MaxWaitTime time.Duration
 	// MaxRetries is the maximum number of retry attempts when waiting for Connect to
 	// propagate changes. The wait function uses exponential backoff between retries.
 	MaxRetries        int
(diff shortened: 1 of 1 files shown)


internal/onepassword/connect/connect_client.go
```diff
@@ -57,22 +72,159 @@ func (c *Client) GetItemByTitle(_ context.Context, title string, vaultUuid strin
...
+		if backoff > 500*time.Millisecond {
+			backoff = 500 * time.Millisecond
+		}
+
+		// Don't sleep on the last attempt
+		if attempt < maxAttempts-1 {
+			time.Sleep(backoff)
+		}
+	}
+	return nil // Max attempts reached, but no error
```
JillRegan: Should we return an error in this case? In `UpdateItem` and `CreateItem` we return the items as if they were successful even on timeout. So subsequent reads may not be able to find the item still. If we add a error return here would it make for easier debugging in the future?

volodymyrZotov: Makes sense. I also removed `MaxWaitTime` to simplify configuration. We'll try `MaxRetries` and return error if condition didn't met and retries loop finishes.

PR #249, opened 2025-11-13
Add new SDK client and conversion helpers

JillRegan:
### ✨ Summary

- Add internal onepassword SDK wrapper and simple models to centralize conversions between the provider and sdk.
- New models: Vault, Item, ItemSection, ItemField, ItemURL, ItemFile, GeneratorRecipe.
- New SDK client with create/read/update/delete item methods, file download, and client initialization (service account or desktop).

### 🔗 Resolves:

- Initial setup for converting terraform provider data to sdk data

### ✅ Checklist

- [x] 🖊️ Commits are signed
  - [ ] 🔹 Unit /🔸 Integration
  - [ ] 🌐 E2E

- NOTE: The SDK setup for the provider is not available for use yet. We are just implementing logic for conversions which will be used in the full SDK implementation in a later PR.

### DIFF
--- a/internal/onepassword/model/item.go
+++ b/internal/onepassword/model/item.go
@@ -204,7 +204,8 @@ func fromSDKFiles(item *sdk.Item, sectionMap map[string]ItemSection) []ItemFile 
 		// Look up section by ID
 		if f.SectionID != "" {
 			if section, exists := sectionMap[f.SectionID]; exists {
-				file.Section = &section
+				file.SectionID = section.ID
+				file.SectionLabel = section.Label
 			}
 		}
 
--- a/internal/onepassword/model/item_file.go
+++ b/internal/onepassword/model/item_file.go
@@ -5,11 +5,12 @@ import (
 )
 
 type ItemFile struct {
-	ID      string
-	Name    string
-	Size    int
-	Section *ItemSection
-	content []byte
+	ID           string
+	Name         string
+	Size         int
+	SectionID    string
+	SectionLabel string
+	content      []byte
 }
 
 func (f *ItemFile) Content() ([]byte, error) {


internal/onepassword/model/item.go
```diff
@@ -0,0 +1,75 @@
...
+type ItemFieldType string
+type FieldPurpose string
+
+const (
+	ItemCategoryLogin         ItemCategory = "Login"
+	ItemCategoryPassword      ItemCategory = "Password"
+	ItemCategoryAPICredential ItemCategory = "ApiCredential"
+	ItemCategoryDatabase      ItemCategory = "Database"
+	ItemCategorySecureNote    ItemCategory = "Secure_Note"
+	ItemCategorySSHKey        ItemCategory = "Ssh_Key"
```
volodymyrZotov: Why do we define own categories? Maybe just re-use from `"github.com/1Password/connect-sdk-go/onepassword"`?

Currently [all the values](https://github.com/1Password/connect-sdk-go/blob/main/onepassword/items.go#L37) are capital case.

internal/onepassword/model/item.go
```diff
@@ -0,0 +1,75 @@
...
+	CharacterSets []string
+}
+
+type ItemURL struct {
+	URL     string
+	Label   string
+	Primary bool
+}
+
+type ItemFile struct {
```
volodymyrZotov: I'd keep this struct in the same file as `func (f *ItemFile) Content() ([]byte, error)` method

internal/onepassword/model/conversions/sdk_conversions.go
```diff
@@ -0,0 +1,146 @@
...
+
+	return field
+}
+
+func toSDKTags(tags []string) []string {
+	if len(tags) == 0 {
+		return nil
+	}
+	return tags
+}
```
volodymyrZotov: Why do we need this check for Tags and do not need it for Sections or Websites?

internal/onepassword/model/conversions/provider_conversions.go
```diff
@@ -0,0 +1,163 @@
+package conversions
```
volodymyrZotov: Do we also need to introduce new file `connect_convertions` and move logic how we map internal provider structs to Connect's?

internal/onepassword/sdk/sdk_client.go
```diff
@@ -0,0 +1,168 @@
...
+
+	content, err := c.sdkClient.Items().Files().Read(ctx, vaultUUID, itemUUID, fileAttributes)
+	if err != nil {
+		return nil, err
+	}
+
+	return content, nil
+}
+
+func NewClient(ctx context.Context, providerUserAgent, serviceAccountToken string, accountToken string) (*Client, error) {
```
volodymyrZotov: let's define `SDKConfig` struct which will have all the properties and pass it as a single `config SDKConfig` argument.

Also I think instead of name `accountToken` more sutable will be `account` or something like that, because we it can be [UUID or name of the account](https://developer.1password.com/docs/sdks/desktop-app-integrations#step-3-connect-your-project-to-the-1password-app).

internal/onepassword/sdk/sdk_client.go
```diff
@@ -0,0 +1,168 @@
...
+	var sdkClient *opsdk.Client
+	var err error
+
+	// Initialize with service account token if provided, otherwise use desktop integration
+	if serviceAccountToken != "" {
+		sdkClient, err = opsdk.NewClient(ctx,
+			opsdk.WithServiceAccountToken(serviceAccountToken),
+			opsdk.WithIntegrationInfo("terraform-provider", providerUserAgent),
```
volodymyrZotov: Currently, we construct user agent value at [the very top](https://github.com/1Password/terraform-provider-onepassword/blob/main/internal/provider/provider.go#L135) and pass it as a config property. We do not need `"terraform-provider"` string hardcoded here, but instead consume the values from `providerUserAgent` parameter.

internal/onepassword/model/conversions/provider_conversions.go
volodymyrZotov: Also how do you feel about `mapper` package name?

internal/onepassword/model/conversions/provider_conversions.go
```diff
@@ -0,0 +1,163 @@
...
+
+import (
+	"strings"
+
+	"github.com/1Password/terraform-provider-onepassword/v2/internal/onepassword/model"
+	sdk "github.com/1password/onepassword-sdk-go"
+)
+
+// FromSDK creates a new Item from an SDK item
+func FromSDKItem(item *sdk.Item) *model.Item {
```
volodymyrZotov: It's not clear from the name to what it returns. What do you think about `FromSDKItemToModel` and then from model to SKD it will be something like `FromModelItemToSDK`?

internal/onepassword/model/conversions/sdk_conversions.go
```diff
@@ -0,0 +1,146 @@
...
+	"context"
+	"fmt"
+
+	"github.com/1Password/terraform-provider-onepassword/v2/internal/onepassword/model"
+	sdk "github.com/1password/onepassword-sdk-go"
+	"github.com/hashicorp/go-uuid"
+)
+
+// ToSDKItem converts internal model Item to SDK ItemCreateParams
+func ToSDKItem(i *model.Item, vaultID string) sdk.ItemCreateParams {
```
volodymyrZotov: - Do we need `vaultID string` parameter? `model.Item` has `VaultID` property?
- Can we name rename this function to clearly indicate that it returns `sdk.ItemCreateParams`?

internal/onepassword/sdk/client.go
```diff
@@ -0,0 +1,168 @@
+package sdk
```
volodymyrZotov, line 1: I think it's fine to name this file as `client`. It's clear that it's sdk client as it is within `sdk` package.

internal/onepassword/model/conversions/provider_conversions.go
volodymyrZotov: Also I think it's fine to drop `_conversions` suffix from file name. As these files already in `conversions` package.

volodymyrZotov: Ignore this, as we decided to move conversion function and made them methods of `model.Item`

internal/onepassword/model/item.go
JillRegan: Updated to use the categories/field types/purposes from connect.

internal/onepassword/model/conversions/sdk_conversions.go
JillRegan: The SDK will return back an empty array when passed an empty array for tags which causes the provider to throw an error, I doesn't look like connect or the cli behaved the same. 

 But I think it is best to handle that when the internal code for the provider so I have removed this for now and can update in a later PR.

internal/onepassword/sdk/sdk_client.go
JillRegan: Done ✅

internal/onepassword/sdk/sdk_client.go
JillRegan: Got it, had to add logic to split the providerUserAgent and the SDK expects two params for `WithIntegrationInfo`.

internal/onepassword/model/conversions/provider_conversions.go
JillRegan: Updated 👍

internal/onepassword/model/conversions/sdk_conversions.go
JillRegan: You're right, removed the vault ID. And updated the function to be more clearly defined.

internal/onepassword/model/item.go
```diff
@@ -0,0 +1,283 @@
...
+	ID       string
+	Title    string
+	VaultID  string
+	Category connect.ItemCategory
+	Version  int
+	Tags     []string
+	URLs     []ItemURL
+	Sections []*ItemSection
+	Fields   []*ItemField
+	Files    []*ItemFile
```
volodymyrZotov: I'd go here without pointers, as this will bring the following benefits:
- Fewer allocations & less GC work. []T stores the elements contiguously in one backing array. []*T needs one allocation per element, which increases GC pressure and pointer chasing.
- Better cache locality. Contiguous structs read/write faster than following pointers all over the heap.
- Simpler semantics. No nil elements to check; you just get zero-values. APIs are easier to use and reason about.
- Copy cost is often negligible. Slices are small headers; passing []T around does not copy the whole array. Even when elements move during append, copying small/medium structs is cheap.

internal/onepassword/sdk/client.go
```diff
@@ -0,0 +1,180 @@
...
+	return v, nil
+}
+
+func (c *Client) GetVaultsByTitle(ctx context.Context, title string) ([]*model.Vault, error) {
+	vaultList, err := c.sdkClient.Vaults().List(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	var result []*model.Vault
```
volodymyrZotov: I'd also define it as `[]model.Vault`. Is there any specific reason to use pointer here?

internal/onepassword/sdk/client.go
```diff
@@ -0,0 +1,180 @@
...
+	for _, vaultOverview := range vaultList {
+		fullVault, err := c.sdkClient.Vaults().GetOverview(ctx, vaultOverview.ID)
```
volodymyrZotov: Do we really need to get an overview?
It seems like both SDK methods returns the same `VaultOverview` struct
```
List(ctx context.Context, params ...VaultListParams) ([]VaultOverview, error)
GetOverview(ctx context.Context, vaultUuid string) (VaultOverview, error)
```

and it's possible to map to `model.Vault` from `vaultList, err := c.sdkClient.Vaults().List(ctx)`

internal/onepassword/sdk/client.go
```diff
@@ -0,0 +1,180 @@
...
+	var count int
+
+	for _, item := range items {
+		if item.Title == title {
+			matchedID = item.ID
+			count++
+		}
+	}
+
+	return c.GetItem(ctx, matchedID, vaultUuid)
```
volodymyrZotov, line 84: We need to throw an error if `count != 1`, to behave the same as Connect [GetItemByTitle](https://github.com/1Password/connect-sdk-go/blob/main/connect/client.go#L305) method.

internal/onepassword/sdk/client.go
```diff
@@ -0,0 +1,180 @@
...
+}
+
+func (c *Client) CreateItem(ctx context.Context, item *model.Item) (*model.Item, error) {
+	params := item.FromModelItemToSDKCreateParams()
+
+	sdkItem, err := c.sdkClient.Items().Create(ctx, params)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create item: %w", err)
```
volodymyrZotov: Maybe update an error message, so it clearly says that SDK is failed? Something like `failed to create item using sdk`.

So it's very easy to distinct between SDK and Connect errors? Though, we might need/want to update Connect errors as well.

internal/onepassword/sdk/client.go
```diff
@@ -0,0 +1,180 @@
...
+	// Convert back to provider model
+	modelItem := &model.Item{}
+	modelItem.FromSDKItemToModel(&updatedItem)
+	return modelItem, nil
+}
+
+func (c *Client) DeleteItem(ctx context.Context, item *model.Item, vaultUuid string) error {
+	err := c.sdkClient.Items().Delete(ctx, vaultUuid, item.ID)
+	if err != nil {
+		return err
```
volodymyrZotov: Maybe mention SDK in error message here as well? `failed to delete item using sdk....`

If we decide to do so, we might need to wrap all the errors in other methods

internal/onepassword/sdk/client.go
```diff
@@ -0,0 +1,180 @@
...
+	}
+
+	return content, nil
+}
+
+func NewClient(ctx context.Context, config SDKConfig) (*Client, error) {
+	var sdkClient *sdk.Client
+	var err error
+
+	integrationName, integrationVersion, _ := strings.Cut(config.ProviderUserAgent, "/")
```
volodymyrZotov: Let's check 3rd param, and if it's `false` let's return an error.

This way we are safe if somebody accidentally change user agent string.

internal/onepassword/sdk/client.go
JillRegan: Thanks for the catch on this, I missed adding it after the loop.

internal/onepassword/model/item.go
JillRegan: Thank you for the context here 👍

internal/onepassword/sdk/client.go
JillRegan: You are correct, GetOverview. was not necessary here.

internal/onepassword/sdk/client.go
JillRegan: I was thinking for performance reasons to create a pointer instead of copying the values on each loop through, but given the little amount of data that is being copied to the struct I think you are right to say to remove the pointer.

internal/onepassword/sdk/client.go
JillRegan: Updated, also added in the other methods (Update, Get, etc). I can update the error messages in connect in the next PR which converts to the new model usage.

internal/onepassword/model/item.go
JillRegan: I have updated Sections / Fields / Files on top level Item to no longer be pointers. 

As for nested section data I have left those as pointers. I believe referencing those as a pointer would be better in this situation in case there are any changes to Sections on the top level item. Example:

```
type ItemFile struct {
	ID      string
	Name    string
	Size    int
	Section *ItemSection
	content []byte
}

type ItemField struct {
	ID       string
	Label    string
	Type     connect.ItemFieldType
	Value    string
	Purpose  connect.ItemFieldPurpose
	Section  *ItemSection
	Recipe   *GeneratorRecipe
	Generate bool
}
```

internal/onepassword/model/item.go
```diff
@@ -0,0 +1,291 @@
...
+type ItemURL struct {
+	URL     string
+	Label   string
+	Primary bool
+}
+
+// FromSDKItemToModel creates a new Item from an SDK item
+func (i *Item) FromSDKItemToModel(item *sdk.Item) {
+	if item == nil {
+		return
```
volodymyrZotov: What do you think about throwing an error in this case? I think that would be better as we will exactly know that items was passed as `nil`, instead of catching the error somewhere down the flow as `i *Item` was not properly populated.

internal/onepassword/model/item.go
```diff
@@ -0,0 +1,291 @@
...
+	Value    string
+	Purpose  connect.ItemFieldPurpose
+	Section  *ItemSection
+	Recipe   *GeneratorRecipe
+	Generate bool
+}
+
+type GeneratorRecipe struct {
+	Length        int
+	CharacterSets []string
```
volodymyrZotov: Seems like the values we can put here are only `"DIGITS"` and `"SYMBOLS"`.

Let's define a dedicated type for those values and use it here instead of `[]string`, aka
```
// Define a custom type for character sets
type CharacterSet string

// Define constants for valid character set values
const (
    CharacterSetDigits  CharacterSet = "DIGITS"
    CharacterSetSymbols CharacterSet = "SYMBOLS"
)
```

internal/onepassword/model/item.go
```diff
@@ -0,0 +1,291 @@
...
+	i.VaultID = item.VaultID
+	i.Category = connect.ItemCategory(item.Category)
+	i.Tags = item.Tags
+	i.URLs = fromSDKURLs(item.Websites)
+
+	// Convert sections/fields/files
+	sectionMap := make(map[string]ItemSection)
+	i.Files = fromSDKFiles(item, sectionMap)
+	i.Sections = fromSDKSections(item, sectionMap)
+	i.Fields = fromSDKFields(item, sectionMap)
```
volodymyrZotov: Seems that `sectionMap` is populated only in `i.Sections = fromSDKSections(item, sectionMap)`.

That means when it executes `i.Files = fromSDKFiles(item, sectionMap)` line `sectionMap` will always be empty.

Do we want to re-order them? Or explicitly populate a map with values, so it's easy to catch such cases?

I'd go with option 2.

internal/onepassword/model/item.go
```diff
@@ -0,0 +1,291 @@
...
+				})
+			}
+		}
+	}
+
+	return fields
+}
+
+func fromSDKFiles(item *sdk.Item, sectionMap map[string]ItemSection) []ItemFile {
+	files := make([]ItemFile, 0, len(item.Files)+1)
```
volodymyrZotov, line 195: Let's add a comment here explaining why we need `+1`

internal/onepassword/model/item.go
```diff
@@ -0,0 +1,291 @@
...
+
+	connect "github.com/1Password/connect-sdk-go/onepassword"
+	sdk "github.com/1password/onepassword-sdk-go"
+)
+
+type Item struct {
+	ID       string
+	Title    string
+	VaultID  string
+	Category connect.ItemCategory
```
volodymyrZotov: Ideally, as we define internal model, I'd like us do not use `connect` types on any internal structs at all, so they be independent. Would it be a big effort to do so?

This is related to other internal types as well.

internal/onepassword/model/item.go
```diff
@@ -0,0 +1,291 @@
...
+	}
+
+	field := sdk.ItemField{
+		ID:        fieldID,
+		Title:     f.Label,
+		FieldType: sdk.ItemFieldType(f.Type),
+		Value:     f.Value,
+	}
+
+	if f.Section.ID != "" {
```
volodymyrZotov: We need to make a `nil` check here, so it doesn't panic.

volodymyrZotov: We need to add `nil` check here so it doesn't panic.

internal/onepassword/model/item.go
JillRegan: Good point, I have updated to throw and error.

`fmt.Errorf("cannot convert nil SDK item to model")`

internal/onepassword/model/item.go
JillRegan: Updated to call on CharacterSet constants.

```
type GeneratorRecipe struct {
	Length        int
	CharacterSets []CharacterSet
}
```

internal/onepassword/model/item.go
JillRegan: Good catch, I have updated to build the section map first as you mentioned with option 2.

internal/onepassword/model/item.go
JillRegan: Done ✅

internal/onepassword/model/item.go
JillRegan: Defined internally. For the following:

- `CharacterSet`
- `ItemCategory`
- `ItemFieldPurpose`
- `ItemFieldType`

internal/onepassword/model/item.go
JillRegan: As discussed this was updated to no longer be a pointer to a section but rather we reference the SectionId and SectionLabel directly on the field:

```
SectionID    string
SectionLabel string
```

volodymyrZotov (approved):
Good job on that! 🔥 Approved!

PR #256, opened 2025-11-19
Remove letter configuration for password recipe

JillRegan:
### ✨ Summary

- Deprecates the letters configuration for password generation as it will no longer be available
- Adds E2E test to ensure letters is not configurable

### 🔗 Resolves:

- N/A deprecates letters configuration

### ✅ Checklist

- [x] 🖊️ Commits are signed
  - [ ] 🔹 Unit /🔸 Integration
  - [x] 🌐 E2E

NOTE - E2E fails still for service account as SDK has not been implemented here

### DIFF
--- a/internal/onepassword/model/item.go
+++ b/internal/onepassword/model/item.go
@@ -16,8 +16,6 @@ type ItemFieldType string
 const (
 	CharacterSetDigits  CharacterSet = "DIGITS"
 	CharacterSetSymbols CharacterSet = "SYMBOLS"
-	// Letters only for use with connect
-	CharacterSetLetters CharacterSet = "LETTERS"
 
 	Login      ItemCategory = "LOGIN"
 	Password   ItemCategory = "PASSWORD"


internal/onepassword/model/item.go
```diff
@@ -510,9 +512,12 @@ func toConnectFields(fields []ItemField) []*connect.ItemField {
 
 		// Include recipe if present
 		if f.Recipe != nil {
-			characterSets := make([]string, len(f.Recipe.CharacterSets))
-			for i, cs := range f.Recipe.CharacterSets {
-				characterSets[i] = string(cs)
```
JillRegan, line 502: Because connect allows for configuration of letters we need to always pass LETTERS to the charset. If we don't letters will be omitted from the password generation.

test/e2e/item_resource_test.go
```diff
@@ -255,6 +253,52 @@ func TestAccItemResourcePasswordGeneration(t *testing.T) {
 	}
 }
 
+// Test that letters is not supported and will error if configured as this field is deprecated
```
JillRegan, line 256: Added new test to ensure attempting to configure letters will result in an error now.

PR #257, opened 2025-11-19
Use SDK client to authenticate using Service Account or User Account

volodymyrZotov:
### ✨ Summary
This PR starts to use SDK client to authenticate using Service Account or User Account.

It also fixes `sdk.GetVaultsBytitle` method by setting `sdk.VaultListParams{DecryptDetails: &decryptDetails}` param to the SDK `Vaults().List()` call, so it returns decrypted vault details.

### ✅ Checklist

- [x] 🖊️ Commits are signed
  - [ ] 🔹 Unit /🔸 Integration
  - [ ] 🌐 E2E

### DIFF
--- a/test/e2e/item_resource_test.go
+++ b/test/e2e/item_resource_test.go
@@ -9,12 +9,12 @@ import (
 	"testing"
 
 	op "github.com/1Password/connect-sdk-go/onepassword"
-	"github.com/1Password/terraform-provider-onepassword/v2/internal/onepassword/model"
 	"github.com/google/uuid"
 	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
 	"github.com/hashicorp/terraform-plugin-testing/terraform"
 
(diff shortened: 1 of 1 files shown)


volodymyrZotov:
```
Error creating 1Password item, got error failed to create item using sdk:
invalid user input: encountered the following errors: An unsupported item
category was specified; field 'username' had an unsupported field type; Only
items with a Login or Password category can contain websites for autofilling
```

Pipeline fails with the issue above, but can be tackled in separate PR.

test/e2e/item_resource_test.go
```diff
@@ -9,6 +9,7 @@ import (
 	"testing"
 
 	op "github.com/1Password/connect-sdk-go/onepassword"
+	"github.com/1Password/terraform-provider-onepassword/v2/internal/onepassword/model"
```
JillRegan: nit: we can move to provider imports section below.

PR #258, opened 2025-11-20
Not throw an error but re-create not existing item

### LINKED ISSUE #97: Recreate item when it is missing
### Your environment

Terraform Provider Version: 1.1.4 (due to [this bug](https://github.com/1Password/terraform-provider-onepassword/issues/96) on later versions)

Connect Server Version: `latest` which at this moment should be [1.7.1](https://hub.docker.com/layers/1password/connect-sync/1.7.1/images/sha256-fde234f8b77aa49df69070a16bc42fc11a6ca3168790816eea27efed652b37c4?context=explore)`

OS: Linux (running the container that 1Password provides)

Terraform Version: v1.4.6

## What happened?
I used the item resource to create an item in a vault. Then I deleted the item in the vault, and ran terraform again. Terraform threw an error after refresh saying that the item wasn't found.

## What did you expect to happen?
I expected that terraform would recreate the item.

## Steps to reproduce
1. Create an item using the provider
2. Delete the item
3. Run terraform again

## Notes & Logs
<!-- Paste any logs here that may help with debugging.
Error message
```
Error: status 404: item evv...y3q not found
```

volodymyrZotov:
### ✨ Summary

Currently, provider throws an fails with the error if item was manually removed from 1Password.

Though, [Terraform recommendation](https://developer.hashicorp.com/terraform/plugin/framework/resources/read#recommendations) is to ignore returning errors that signify the resource is no longer existent, call the response state `RemoveResource()` method, and return early. The next Terraform plan will recreate the resource.

This PR fixes this issue.

### 🔗 Resolves:
Resolves #97 

### ✅ Checklist

- [x] 🖊️ Commits are signed
  - [ ] 🔹 Unit /🔸 Integration
  - [x] 🌐 E2E

There is e2e tests added that covers exactly this case, in case if want to test manually:

#### Manually testing steps:
1. Checkout to this branch, `make build` and configure provider to use local build in `~/.terraformrc`
2. Define simple item resource in your terraform configuration.
3. `terraform apply` to create an item in 1Password
4. Manually remove item from 1Password vault (via UI or using CLI)
5. Run `terraform apply` again -> it should not error but create an item.

### DIFF
--- a/test/e2e/item_resource_test.go
+++ b/test/e2e/item_resource_test.go
@@ -1,8 +1,10 @@
 package integration
 
 import (
+	"context"
 	"fmt"
 	"maps"
+	"os"
 	"regexp"
 	"testing"
 
@@ -11,6 +13,7 @@ import (
 	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
 	"github.com/hashicorp/terraform-plugin-testing/terraform"
 
+	"github.com/1Password/terraform-provider-onepassword/v2/internal/onepassword"
 	tfconfig "github.com/1Password/terraform-provider-onepassword/v2/test/e2e/terraform/config"
 	"github.com/1Password/terraform-provider-onepassword/v2/test/e2e/utils/checks"
 	"github.com/1Password/terraform-provider-onepassword/v2/test/e2e/utils/password"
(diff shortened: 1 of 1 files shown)


JillRegan (approved):
Thanks for the fix on this! ⭐

PR #262, opened 2025-11-25
Add account e2e tests

JillRegan:
### ✨ Summary

- Adds E2E test TestAccItemResourceWithAccount which uses desktop auth to get items. 
- This test can only be run locally and user needs to create a vault called Terraform Test Vault in order to run

### 🔗 Resolves:

- Creates test for desktop auth with SDK

### ✅ Checklist

- [x] 🖊️ Commits are signed
  - [ ] 🔹 Unit /🔸 Integration
  - [ ] 🌐 E2E

- Can only be tested locally, is skipped in the pipeline

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -33,8 +33,9 @@ test-e2e-connect: ## Run e2e tests using Connect (requires OP_CONNECT_TOKEN and 
 	@echo "[INFO] Running e2e tests with Connect authentication..."
 	@sh -c 'unset OP_SERVICE_ACCOUNT_TOKEN; OP_CONNECT_TOKEN="$(OP_CONNECT_TOKEN)" OP_CONNECT_HOST="$(OP_CONNECT_HOST)" TF_ACC=1 go test -v ./test/e2e/... -timeout 30m'
 
-test-e2e-account: ## Run e2e test using account methodology with Touch ID (requires OP_ACCOUNT, manual only). Creates multiple items to verify biometrics is only prompted once.
+test-e2e-account: ## Run e2e test using account methodology with Touch ID (requires OP_ACCOUNT and OP_TEST_VAULT_NAME, manual only). Creates multiple items to verify biometrics is only prompted once.
 	@test -n "$(OP_ACCOUNT)" || (echo "[ERROR] OP_ACCOUNT environment variable is not set."; exit 1)
+	@test -n "$(OP_TEST_VAULT_NAME)" || (echo "[ERROR] OP_TEST_VAULT_NAME environment variable is not set."; exit 1)
 	@echo "[INFO] Running e2e test with account-based authentication (Touch ID)..."
 	@echo "[WARNING] This test will prompt for Touch ID/biometric authentication."
 	@echo "[WARNING] Please ensure you add a vault titled 'terraform-provider-acceptance-tests' or set OP_TEST_VAULT_NAME environment variable for this test to pass."


test/e2e/terraform/config/item_resource_config.go
```diff
@@ -6,11 +6,24 @@ import (
...
+}
+
+func ItemResourceConfigWithName(resourceName string, vaultID string, params map[string]any) func() string {
 	return func() string {
-		resourceStr := `resource "onepassword_item" "test_item" {`
+		resourceStr := fmt.Sprintf(`resource "onepassword_item" %q {`, resourceName)
 
-		resourceStr += fmt.Sprintf("\n  vault = %q", vaultID)
+		// Terraform references must not be quoated
+		if strings.HasPrefix(vaultID, "data.") {
```
volodymyrZotov: In which case we provide `data.` as vault ID? Maybe need to fix the place where it  happens so it always pass vaild uuid without `data.` prefix?

Makefile
```diff
@@ -33,6 +33,14 @@ test-e2e-connect: ## Run e2e tests using Connect (requires OP_CONNECT_TOKEN and
 	@echo "[INFO] Running e2e tests with Connect authentication..."
 	@sh -c 'unset OP_SERVICE_ACCOUNT_TOKEN; OP_CONNECT_TOKEN="$(OP_CONNECT_TOKEN)" OP_CONNECT_HOST="$(OP_CONNECT_HOST)" TF_ACC=1 go test -v ./test/e2e/... -timeout 30m'
 
+test-e2e-account: ## Run e2e test using account methodology with Touch ID (requires OP_ACCOUNT, manual only). Creates multiple items to verify biometrics is only prompted once.
+	@test -n "$(OP_ACCOUNT)" || (echo "[ERROR] OP_ACCOUNT environment variable is not set."; exit 1)
+	@echo "[INFO] Running e2e test with account-based authentication (Touch ID)..."
+	@echo "[WARNING] This test will prompt for Touch ID/biometric authentication."
+	@echo "[WARNING] Please ensure you add a vault titled 'Terraform Test Vault' for this test to pass."
+	@echo "[INFO] The test creates multiple items - observe that biometrics is only prompted once."
+	@sh -c 'unset OP_CONNECT_TOKEN OP_CONNECT_HOST OP_SERVICE_ACCOUNT_TOKEN; OP_ACCOUNT="$(OP_ACCOUNT)" TF_ACC=1 go test -v ./test/e2e/... -run TestAccItemResourceWithAccount -timeout 30m'
```
volodymyrZotov: Why do we run only single test? Can we run all e2e as we do when using Connect or SA?

JillRegan: I have updated this in a way to just run all resource tests by creating a helper to query to get the vault uuid by name, that way any user can create a test vault and use it by setting OP_TEST_VAULT_NAME, or we default back to the terraform-provider-acceptance-tests vault. 

The data source tests are problematic though as we get them by uuid, which is specific to those in the original test vault.

test/e2e/utils/vault/vault.go
```diff
@@ -0,0 +1,46 @@
...
+	"context"
+	"os"
+	"sync"
+	"testing"
+
+	"github.com/1Password/terraform-provider-onepassword/v2/test/e2e/utils/client"
+)
+
+var (
+	testVaultIDOnce sync.Once
```
JillRegan, line 13: Using Once from the [sync package](https://pkg.go.dev/sync#Once) means we can just execute this method once and cache the result to ensure we don't read the vault for every test case.

test/e2e/terraform/config/item_resource_config.go
JillRegan: Was able to remove this with the update to just converting to running resource tests with account.

test/e2e/utils/vault/vault.go
JillRegan: We will also need this helper as we need to get the uuid for the vault we are running the tests for. We cannot use vault name with resource items, only [vault uuid](https://registry.terraform.io/providers/1Password/onepassword/latest/docs/resources/item).

volodymyrZotov: I love this improvement! 🔥

volodymyrZotov:
> This test can only be run locally and user needs to create a vault called Terraform Test Vault in order to run

Could you mention this in `docs/testing/testing.md`, please? 

Can be done in a followup PR, as I we also need to update there instructions about setting Connect env vars.

Makefile
```diff
@@ -33,6 +33,14 @@ test-e2e-connect: ## Run e2e tests using Connect (requires OP_CONNECT_TOKEN and
 	@echo "[INFO] Running e2e tests with Connect authentication..."
 	@sh -c 'unset OP_SERVICE_ACCOUNT_TOKEN; OP_CONNECT_TOKEN="$(OP_CONNECT_TOKEN)" OP_CONNECT_HOST="$(OP_CONNECT_HOST)" TF_ACC=1 go test -v ./test/e2e/... -timeout 30m'
 
+test-e2e-account: ## Run e2e test using account methodology with Touch ID (requires OP_ACCOUNT, manual only). Creates multiple items to verify biometrics is only prompted once.
+	@test -n "$(OP_ACCOUNT)" || (echo "[ERROR] OP_ACCOUNT environment variable is not set."; exit 1)
```
volodymyrZotov, line 37: Let's add the same check for `OP_TEST_VAULT_NAME` env var, so it doesn't run tests if it's not set.

volodymyrZotov (approved):

PR #264, opened 2025-11-28
Deprecate cli code

JillRegan:
### ✨ Summary

- This PR deprecates all related CLI code for the provider as we are moving to SDKs.
- Update to docs for SDK code will come later

### ✅ Checklist

- [x] 🖊️ Commits are signed
- [ ] 🧪 Tests added/updated: _(See the [Testing Guide](docs/testing/testing.md) for when to use each type and how to run them)_
  - [ ] 🔹 Unit /🔸 Integration
  - [ ] 🌐 E2E
- [x] 📚 Docs updated (if behavior changed)

PR #265, opened 2025-12-01
Update auth params

JillRegan:
### ✨ Summary

- This PR updated params `url` and `token` to be more specific to Connect. Updated to `connect_url` and `connect_token`.

### ✅ Checklist

- [x] 🖊️ Commits are signed
  - [ ] 🔹 Unit /🔸 Integration
  - [ ] 🌐 E2E

### DIFF
--- a/internal/onepassword/client.go
+++ b/internal/onepassword/client.go
@@ -41,5 +41,5 @@ func NewClient(ctx context.Context, config ClientConfig) (Client, error) {
 			ProviderUserAgent: config.ProviderUserAgent,
 		}), nil
 	}
-	return nil, errors.New("Invalid provider configuration. Either Connect credentials (\"connect_token\" and \"connect_url\") or Service Account (\"service_account_token\" or \"account\") credentials should be set.")
+	return nil, errors.New("Invalid provider configuration. Either Connect credentials (\"connect_token\" and \"connect_url\") or Service Account (\"service_account_token\") or \"account\"  should be set.")
 }
--- a/internal/provider/provider.go
+++ b/internal/provider/provider.go
@@ -52,25 +52,11 @@ func (p *OnePasswordProvider) Schema(ctx context.Context, req provider.SchemaReq
 			"connect_url": schema.StringAttribute{
 				MarkdownDescription: "The HTTP(S) URL where your 1Password Connect server can be found. Can also be sourced `OP_CONNECT_HOST` environment variable. Provider will use 1Password Connect server if set.",
 				Optional:            true,
-				Validators: []validator.String{
-					stringvalidator.ConflictsWith(
-						path.Expressions{
-							path.MatchRoot("url"),
-						}...,
-					),
-				},
 			},
 			"connect_token": schema.StringAttribute{
 				MarkdownDescription: "A valid token for your 1Password Connect server. Can also be sourced from `OP_CONNECT_TOKEN` environment variable. Provider will use 1Password Connect server if set.",
 				Optional:            true,
 				Sensitive:           true,
-				Validators: []validator.String{
-					stringvalidator.ConflictsWith(
-						path.Expressions{
-							path.MatchRoot("token"),
-						}...,
-					),
-				},
 			},
 			"url": schema.StringAttribute{
 				MarkdownDescription: "The HTTP(S) URL where your 1Password Connect server can be found. Can also be sourced `OP_CONNECT_HOST` environment variable. Provider will use 1Password Connect server if set. Deprecated: Use `connect_url` instead.",
@@ -82,6 +68,7 @@ func (p *OnePasswordProvider) Schema(ctx context.Context, req provider.SchemaReq
 						}...,
 					),
 				},
+				DeprecationMessage: "The \"url\" field is deprecated and will be removed in a future version. Use \"connect_url\" instead.",
 			},
 			"token": schema.StringAttribute{
 				MarkdownDescription: "A valid token for your 1Password Connect server. Can also be sourced from `OP_CONNECT_TOKEN` environment variable. Provider will use 1Password Connect server if set. Deprecated: Use `connect_token` instead.",
@@ -94,6 +81,7 @@ func (p *OnePasswordProvider) Schema(ctx context.Context, req provider.SchemaReq
 						}...,
 					),
 				},
+				DeprecationMessage: "The \"token\" field is deprecated and will be removed in a future version. Use \"connect_token\" instead.",
 			},
 			"service_account_token": schema.StringAttribute{
 				MarkdownDescription: "A valid 1Password service account token. Can also be sourced from `OP_SERVICE_ACCOUNT_TOKEN` environment variable.",
@@ -134,11 +122,9 @@ func (p *OnePasswordProvider) Configure(ctx context.Context, req provider.Config
 
 	// Old field names - these are deprecated and will be removed in a future version.
 	if !config.ConnectHostOld.IsNull() {
-		resp.Diagnostics.AddWarning("Deprecated field", "The \"url\" field is deprecated and will be removed in a future version. Use \"connect_url\" instead.")
 		connectHost = config.ConnectHostOld.ValueString()
 	}
 	if !config.ConnectTokenOld.IsNull() {
-		resp.Diagnostics.AddWarning("Deprecated field", "The \"token\" field is deprecated and will be removed in a future version. Use \"connect_token\" instead.")
 		connectToken = config.ConnectTokenOld.ValueString()
 	}
 
--- a/test/e2e/item_resource_test.go
+++ b/test/e2e/item_resource_test.go
@@ -554,8 +554,8 @@ func TestAccItemResourceTags(t *testing.T) {
 		tags []string
 	}{
 		{"CREATE_ITEM_WITH_2_TAGS", []string{"firstTestTag", "secondTestTag"}},
-		// {"ADD_3RD_TAG", []string{"firstTestTag", "secondTestTag", "thirdTestTag"}},
-		// {"REMOVE_2_TAGS", []string{"firstTestTag"}},
+		{"ADD_3RD_TAG", []string{"firstTestTag", "secondTestTag", "thirdTestTag"}},
+		{"REMOVE_2_TAGS", []string{"firstTestTag"}},
 	}
 
 	testVaultID := vault.GetTestVaultID(t)


internal/provider/provider.go
```diff
@@ -29,8 +29,8 @@ type OnePasswordProvider struct {
 
 // OnePasswordProviderModel describes the provider data model.
 type OnePasswordProviderModel struct {
-	ConnectHost         types.String `tfsdk:"url"`
-	ConnectToken        types.String `tfsdk:"token"`
+	ConnectHost         types.String `tfsdk:"connect_url"`
+	ConnectToken        types.String `tfsdk:"connect_token"`
```
volodymyrZotov, line 36: I though about this a bit more.

What if we still keep `url` and `token` config properties available, but mark them as deprecated, so it's easier for the users to adopt a new version and we can completely deprecate it in the next release.

```
// New field names (preferred)
	ConnectHost         types.String `tfsdk:"connect_url"`
	ConnectToken        types.String `tfsdk:"connect_token"`
	// Old field names (deprecated, for backward compatibility)
	ConnectHostOld      types.String `tfsdk:"url"`
	ConnectTokenOld     types.String `tfsdk:"token"`
```

internal/provider/provider.go
```diff
@@ -43,11 +43,11 @@ func (p *OnePasswordProvider) Metadata(ctx context.Context, req provider.Metadat
 func (p *OnePasswordProvider) Schema(ctx context.Context, req provider.SchemaRequest, resp *provider.SchemaResponse) {
 	resp.Schema = schema.Schema{
 		Attributes: map[string]schema.Attribute{
-			"url": schema.StringAttribute{
+			"connect_url": schema.StringAttribute{
 				MarkdownDescription: "The HTTP(S) URL where your 1Password Connect server can be found. Can also be sourced `OP_CONNECT_HOST` environment variable. Provider will use 1Password Connect server if set.",
 				Optional:            true,
 			},
```
volodymyrZotov, line 55: If we go with keeping both `connect_url` and `url` params, we can add validation using framework validation package to make sure only one of them is set (either `url` or `connect_url`) aka
```
Validators: []validator.String{
					stringvalidator.ConflictsWith(
						path.Expressions{
							path.MatchRoot("url"),
						}...,
					),
				},
```
The same is applicable for `token`

internal/provider/provider.go
JillRegan: Good point, I have updated to allow use of these fields 👍

docs/index.md
```diff
@@ -44,9 +44,11 @@ Environment variables are a more secure way to set configuration options because
 ### Optional
 
 - `account` (String) A valid account name or ID to use biometrics unlock. Can also be sourced from `OP_ACCOUNT` environment variable. Provider will use the 1Password SDK if set.
+- `connect_token` (String, Sensitive) A valid token for your 1Password Connect server. Can also be sourced from `OP_CONNECT_TOKEN` environment variable. Provider will use 1Password Connect server if set.
+- `connect_url` (String) The HTTP(S) URL where your 1Password Connect server can be found. Can also be sourced `OP_CONNECT_HOST` environment variable. Provider will use 1Password Connect server if set.
 - `service_account_token` (String, Sensitive) A valid 1Password service account token. Can also be sourced from `OP_SERVICE_ACCOUNT_TOKEN` environment variable. Provider will use the 1Password SDK if set.
```
volodymyrZotov:
```suggestion
- `service_account_token` (String, Sensitive) A valid 1Password service account token. Can also be sourced from `OP_SERVICE_ACCOUNT_TOKEN` environment variable.
```

internal/provider/provider.go
```diff
@@ -104,7 +158,7 @@ func (p *OnePasswordProvider) Configure(ctx context.Context, req provider.Config
 	// TODO: Investigate if wrapping this as a (framework) validator can be a better fit.
```
volodymyrZotov: We can remove this todo.

I looked into that, and we could not compleatly switch to framework validator, as it validates only input params aka `connect_host`/`connect_token` provided in `.tf`. But it could not validate values provided as env vars. Therefore we still need this hand written validation

internal/provider/provider.go
```diff
@@ -104,7 +158,7 @@ func (p *OnePasswordProvider) Configure(ctx context.Context, req provider.Config
...
 	if serviceAccountToken != "" || account != "" {
 		if connectToken != "" || connectHost != "" {
-			resp.Diagnostics.AddError("Config conflict", "Either Connect credentials (\"token\" and \"url\") or 1Password SDK (\"service_account_token\" or \"account\") credentials can be set. Both are set. Please unset one of them.")
+			resp.Diagnostics.AddError("Config conflict", "Either Connect credentials (\"connect_token\" and \"connect_url\") or 1Password SDK (\"service_account_token\" or \"account\") credentials can be set. Both are set. Please unset one of them.")
```
volodymyrZotov: This will be addressed in this [PR](https://github.com/1Password/terraform-provider-onepassword/pull/264/files#diff-58d6a027753b50994deb7e11e4a99dde423f35844986019bd9cea5e0c94aba22R107).

Maybe update it here to be the same to not overwrite if merging that PR first and this one second.

internal/provider/provider.go
```diff
@@ -104,7 +158,7 @@ func (p *OnePasswordProvider) Configure(ctx context.Context, req provider.Config
...
 		}
 		if serviceAccountToken != "" && account != "" {
 			resp.Diagnostics.AddError("Config conflict", "\"service_account_token\" and \"account\" are set. Please unset one of them to use the provider with 1Password SDK.")
```
volodymyrZotov: The same here. Merge this PR first and then [another one](https://github.com/1Password/terraform-provider-onepassword/pull/264/files#diff-58d6a027753b50994deb7e11e4a99dde423f35844986019bd9cea5e0c94aba22) so it removes mention about SDK

internal/onepassword/client.go
```diff
@@ -41,5 +41,5 @@ func NewClient(ctx context.Context, config ClientConfig) (Client, error) {
 			ProviderUserAgent: config.ProviderUserAgent,
 		}), nil
 	}
-	return nil, errors.New("Invalid provider configuration. Either Connect credentials (\"token\" and \"url\") or Service Account (\"service_account_token\" or \"account\") credentials should be set.")
+	return nil, errors.New("Invalid provider configuration. Either Connect credentials (\"connect_token\" and \"connect_url\") or Service Account (\"service_account_token\" or \"account\") credentials should be set.")
```
volodymyrZotov:
```suggestion
	return nil, errors.New("Invalid provider configuration. Either Connect credentials (\"connect_token\" and \"connect_url\") or Service Account (\"service_account_token\") or \"account\"  should be set.")
```

internal/provider/provider.go
```diff
@@ -43,14 +49,51 @@ func (p *OnePasswordProvider) Metadata(ctx context.Context, req provider.Metadat
...
 				Optional:            true,
 				Sensitive:           true,
+				Validators: []validator.String{
+					stringvalidator.ConflictsWith(
+						path.Expressions{
+							path.MatchRoot("token"),
+						}...,
+					),
+				},
+			},
```
volodymyrZotov: `[nit]`
It's enough to have only one validator set for `token` parameter.

If set validators for both it prints 2 errors:
```
Attribute "connect_token" cannot be specified when "token" is specified
 Attribute "token" cannot be specified when "connect_token" is specified
```

But if we have only one for `token`, it will print 1 error:
```
Attribute "connect_token" cannot be specified when "token" is specified
```

The same is aplicable for `url`

internal/provider/provider.go
```diff
@@ -43,14 +49,51 @@ func (p *OnePasswordProvider) Metadata(ctx context.Context, req provider.Metadat
...
+						path.Expressions{
+							path.MatchRoot("connect_url"),
+						}...,
+					),
+				},
+			},
+			"token": schema.StringAttribute{
+				MarkdownDescription: "A valid token for your 1Password Connect server. Can also be sourced from `OP_CONNECT_TOKEN` environment variable. Provider will use 1Password Connect server if set. Deprecated: Use `connect_token` instead.",
+				Optional:            true,
+				Sensitive:           true,
```
volodymyrZotov, line 76: Depreciation message can be set as a property
```
DeprecationMessage:  "depriocation message for token",
```
So, `resp.Diagnostics.AddWarning("Deprecated field", "The \"url\" field is deprecated and will be removed in a future version. Use \"connect_url\" instead.")` and `resp.Diagnostics.AddWarning("Deprecated field", "The \"token\" field is deprecated and will be removed in a future version. Use \"connect_token\" instead.")` which are added manually, can be removed.

It will print warning like that:

Image

internal/provider/provider.go
```diff
@@ -88,6 +131,17 @@ func (p *OnePasswordProvider) Configure(ctx context.Context, req provider.Config
 	if !config.ConnectToken.IsNull() {
 		connectToken = config.ConnectToken.ValueString()
 	}
+
+	// Old field names - these are deprecated and will be removed in a future version.
+	if !config.ConnectHostOld.IsNull() {
+		resp.Diagnostics.AddWarning("Deprecated field", "The \"url\" field is deprecated and will be removed in a future version. Use \"connect_url\" instead.")
```
volodymyrZotov: can be removed as per https://github.com/1Password/terraform-provider-onepassword/pull/265/files#r2582753301

internal/provider/provider.go
```diff
@@ -88,6 +131,17 @@ func (p *OnePasswordProvider) Configure(ctx context.Context, req provider.Config
...
+		connectHost = config.ConnectHostOld.ValueString()
+	}
+	if !config.ConnectTokenOld.IsNull() {
+		resp.Diagnostics.AddWarning("Deprecated field", "The \"token\" field is deprecated and will be removed in a future version. Use \"connect_token\" instead.")
```
volodymyrZotov: can be removed as per comment https://github.com/1Password/terraform-provider-onepassword/pull/265/files#r2582753301

test/e2e/item_resource_test.go
```diff
@@ -554,8 +554,8 @@ func TestAccItemResourceTags(t *testing.T) {
 		tags []string
 	}{
 		{"CREATE_ITEM_WITH_2_TAGS", []string{"firstTestTag", "secondTestTag"}},
-		// {"ADD_3RD_TAG", []string{"firstTestTag", "secondTestTag", "thirdTestTag"}},
-		// {"REMOVE_2_TAGS", []string{"firstTestTag"}},
```
JillRegan, line 558: Noticed these were commented out, must have happened in a previous PR when I was testing and got missed.

volodymyrZotov (approved):

PR #266, opened 2025-12-01
 Add new SDK docs

JillRegan:
### ✨ Summary

-Add documentation for new SDK implementation

### ✅ Checklist

- [x] 🖊️ Commits are signed
  - [ ] 🔹 Unit /🔸 Integration
  - [ ] 🌐 E2E

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -12,16 +12,20 @@ assignees: ""
 
 Terraform Provider Version:
 
-<!-- What version of the Connect server are you running?
+<!-- What authentication method are you using?
+- Connect: Using Connect server credentials (connect_url + connect_token)
+- Service Account: Using service account token
+- Desktop app: Using account name/ID for biometric unlock
+-->
+
+Authentication Method:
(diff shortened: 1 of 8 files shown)


docs/index.md
```diff
@@ -12,6 +12,34 @@ Use the 1Password Terraform provider to reference, create, or update items in yo
 The 1Password Terraform provider supports using both [1Password Connect Server](https://developer.1password.com/docs/secrets-automation/#1password-connect-server)
 and [1Password SDK](https://developer.1password.com/docs/sdks/).
 
+## Use with 1Password SDK
```
volodymyrZotov: Does it important for that users to know what library (in our case 1password SDK) provider uses under the hood? In my opinion that doesn't really matters, as that may change. What matters more is which credentials you can use to authenticate and access secrets.

What if we say it like:
1. Authenticate with Service Account
2. Authenticate with biometric unlock
3. Authenticate with Connect credentials

How does it sound?

JillRegan: Probably not no, I think it makes sense to remove specific reference to what libraries we use. I have updated this to list authentication options like you mentioned.

PR #281, opened 2025-12-04
Fix release process

volodymyrZotov:
### ✨ Summary

This PR fixes release process which is broken as of started using `onpassword-sdk`, as sdk is built with CGO enabled.

The changes in this PR are:
1. Enables build and release with CGO
2. Creates separate release workflow for each platform darwin/linux/windows
3. Creates separate goreleaser config for each platform
4. Bumps go version to `v1.25.5` for the project
5. Uses the go version specific in `go.mod` in all workflows

### ✅ Checklist

- [x] 🖊️ Commits are signed
  - [ ] 🔹 Unit /🔸 Integration
  - [ ] 🌐 E2E

### DIFF
--- a/.github/workflows/release-darwin.yml
+++ b/.github/workflows/release-darwin.yml
@@ -34,7 +34,7 @@ jobs:
       - name: Run GoReleaser
         uses: goreleaser/goreleaser-action@v6
         with:
-          args: release --clean
+          args: release --clean --config .goreleaser.darwin.yml
         env:
           GPG_FINGERPRINT: ${{ steps.import_gpg.outputs.fingerprint }}
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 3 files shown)


.goreleaser.windows.yml
```diff
@@ -0,0 +1,53 @@
...
+      - "--output"
+      - "${signature}"
+      - "--detach-sign"
+      - "${artifact}"
+release:
+  extra_files:
+    - glob: "terraform-registry-manifest.json"
+      name_template: "{{ .ProjectName }}_{{ .Version }}_manifest.json"
+  # If you want to manually examine the release before its live, uncomment this line:
+  draft: true
```
JillRegan, line 51: Are you wanting to have this commented out? Same with the other two. This will set a manual release won't it?

volodymyrZotov: Yes, we want to keep `draft: true` enabled. This creates draft releases that require manual publishing, which is intentional for our release process. This way we keep it as it always was. Check out `.goreleaser.yaml` has it enabled too.

Issue #287, opened 2025-12-08
API credential items cannot be read with v3


michemache:
### Your environment

Terraform Provider Version: v3.0.0

Authentication Method: Service Account and Desktop app

Connect Server Version: -

OS: macOS 26

Terraform Version: OpenTofu v1.10.7

## What happened?

The credential value of an item with the category  "API credential"  read with the data source `onepassword_item` are `null` after upgrading to v3.

## What did you expect to happen?

The credential value is not `null`, as was previously the case with provider v2.

## Steps to reproduce

1. Create a new API credentials item in your test vault
2. Read the new item with data source `onepassword_item`
```
data "onepassword_vault" "this" {
  name = "test"
}

data "onepassword_item" "this" {
  vault = data.onepassword_vault.this.uuid
  title = "category-api-credential-item"
}

output "item" {
  value     = data.onepassword_item.this
  sensitive = true
}
```
3. Check the output of the data source
```
% tofu output item
{
  "category" = ""
  "credential" = tostring(null) <-- should not be empty
  "database" = tostring(null)
  "file" = tolist([])
  "hostname" = ""
  "id" = "vaults/<vault_id>/items/<itme_id>"
  "note_value" = tostring(null)
  "password" = tostring(null)
  "port" = tostring(null)
  "private_key" = tostring(null)
  "private_key_openssh" = tostring(null)
  "public_key" = tostring(null)
  "section" = tolist([])
  "tags" = tolist([])
  "title" = "category-api-credential-item"
  "type" = "test"
  "url" = tostring(null)
  "username" = "test"
  "uuid" = "<itme_id>"
  "vault" = "<vault_id>"
}

```
## Notes & Logs
The same configuration works without any problems before the provider update to v3.

JillRegan:
The fix for this has been merged, I will update this issue once we have released it. 👍

michemache:
@JillRegan Thanks, are there any plans to release a patch release soon?

JillRegan:
@michemache We are targeting a release for the beginning of next week 👍

JillRegan:
Fix released with v3.0.1

PR #290, opened 2025-12-08
Add API Credential logic to model conversions

JillRegan:
### ✨ Summary

The category comparison was failing to set the API Credential Category because in new item conversions whic hwere implemented with the migration to SDK we did not handle for this category as support here is not official. This type comparison issue prevented the credential field from being populated.

I have added a API Credential category and added it to the mapping helpers. I also moved the field setting logic for `credential` to the field switch case. E2E test has been added.

### 🔗 Resolves:

Resolves: https://github.com/1Password/terraform-provider-onepassword/issues/287

### ✅ Checklist

- [x] 🖊️ Commits are signed
  - [ ] 🔹 Unit /🔸 Integration
  - [x] 🌐 E2E

The bug can reproduced manually using v3.0.0 by attempting to access the credential field in an API Credential Item. The value ends up being null as we never set the category on the item. 

```
data "onepassword_item" "api_credential" {
  vault = data.onepassword_vault.test.uuid
  title = "Jill Test Api Credential"
}

# The bug: credential should contain the API credential value but it's null (also no Category set)
output "api_credential_item" {
  value     = data.onepassword_item.api_credential
  sensitive = true
}

```

### DIFF
--- a/internal/onepassword/model/item.go
+++ b/internal/onepassword/model/item.go
@@ -429,7 +429,6 @@ func fromConnectFields(fields []*connect.ItemField, sectionMap map[string]ItemSe
 		// Connect returns dates as timestamp
 		// Converting timestamp to `YYYY-MM-DD` string.
 		if f.Type == connect.FieldTypeDate && field.Value != "" {
-			fmt.Printf("DEBUG: Processing date field - ID: %q, Label: %q, Value: %q\n", field.ID, field.Label, field.Value)
 			dateStr, err := util.SecondsToYYYYMMDD(field.Value)
 			if err != nil {
 				return modelFields, fmt.Errorf("fromConnectFields: failed to parse timestamp %s to 'YYYY-MM-DD' string format: %w", field.Value, err)


internal/onepassword/model/item.go
```diff
@@ -427,7 +428,7 @@ func fromConnectFields(fields []*connect.ItemField, sectionMap map[string]ItemSe
 		// Provider handles dates in `YYYY-MM-DD` format.
 		// Connect returns dates as timestamp
 		// Converting timestamp to `YYYY-MM-DD` string.
-		if f.Type == connect.FieldTypeDate {
+		if f.Type == connect.FieldTypeDate && field.Value != "" {
```
JillRegan, line 431: There are `validFrom` and `expires` date fields returned from Connect/SDK which were empty strings when set, causing an error here. 

I think it would be good to handle for these fields but I want to address in a separate PR. Currently we don't "officially" support API Credentials with the provider, but many users are using it with this item. 

I would like to add official support for this item and its fields,  as well add documentation about it. This should be handled separately as it would be an enhancement. There is a related [PR](https://github.com/1Password/terraform-provider-onepassword/pull/210) open for this as well.

Issue #274, opened 2025-12-03
Ensure E2E tests destroy all created test data at the end of run


JillRegan:
### Your environment

Terraform Provider Version:

2.2.1

Connect Server Version:

CLI Version:

OS:

Terraform Version:

## What happened?

When E2E test fail in the pipeline due to conflict error, the teardown does not run and items do not get destroyed resulting in abandoned test data cluttering up the test vault. 

## What did you expect to happen?
Even if there is a failure/error all items that are created during E2E tests should be destroyed at the end of the test runs. The only items that should remain in this vault are the ones specific to the data source read tests.

## Steps to reproduce
1. Run E2E in several parallel pipelines which trigger a conflict error:  

```
status 400: Network: (DB: (409) (Conflict), Internal server conflict.), DB:
        (409) (Conflict), Internal server conflict.
```
2. View the terraform-acceptance-tests-vault and see that the items that had the failure have not been destroyed and remain abandoned in the vault.

volodymyrZotov:
`(409) (Conflict)` error happens because of `vault.version` mismatch as we run multiple e2e tests pipelines in parallel that uses the same vault.

Workaround for that could be, to create separate vault for each workflow and delete it after the test finishes.
With that we make sure no stale items are left if pipeline fails.

PR #294, opened 2025-12-15
Add test cleanup helper

JillRegan:
### ✨ Summary

- Certain cases where a 409 / 500 error occur from 1Password the test fails and items are left abandoned in the test vault.
- This PR adds a test clean up method which executes if an E2E test fails which captures the UUID and registers clean up.

Note we already have retries for 409s in the provider. But DELETE unfortunately returns generic 500 response on 409 conflict. This should cleanup issues where item is abandoned on DELETE error.

### 🔗 Resolves:

Resolves: https://github.com/1Password/terraform-provider-onepassword/issues/274

### ✅ Checklist

- [x] 🖊️ Commits are signed
  - [ ] 🔹 Unit /🔸 Integration
  - [x] 🌐 E2E

To test this change checkout the branch locally to run tests as well as rerunning the E2E workflow on this pipeline. There will be sporadic 500 failures but no items should be left abandoned. 

I tested running the job and tests locally several times with failures to ensure all items are cleaned up.

### DIFF
--- a/test/e2e/utils/cleanup/cleanup.go
+++ b/test/e2e/utils/cleanup/cleanup.go
@@ -2,42 +2,55 @@ package cleanup
 
 import (
 	"context"
+	"fmt"
 	"testing"
 	"time"
+
+	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
+	"github.com/hashicorp/terraform-plugin-testing/terraform"
 
 	"github.com/1Password/terraform-provider-onepassword/v2/internal/onepassword/model"
 	"github.com/1Password/terraform-provider-onepassword/v2/test/e2e/utils/client"
 )
 
-// RegisterItemCleanup registers a cleanup function that will delete an item even if the test fails.
-func RegisterItemCleanup(t *testing.T, itemUUID, vaultID string) {
-	if itemUUID == "" {
-		return
+// RegisterItem registers a cleanup function that will delete an item even if the test fails
+func RegisterItem(t *testing.T, uuidPtr *string, vaultID string) resource.TestCheckFunc {
+	return func(s *terraform.State) error {
+		if uuidPtr == nil || *uuidPtr == "" {
+			return fmt.Errorf("cleanup: item UUID is empty")
+		}
+
+		itemUUID := *uuidPtr
+		t.Cleanup(func() {
+			// Only delete if test failed because if test succeeded Terraform already deleted it
+			if !t.Failed() {
+				return
+			}
+
+			ctx := context.Background()
+			testClient, err := client.CreateTestClient(ctx)
+			if err != nil {
+				t.Logf("Cleanup: failed to create client for item %s: %v", itemUUID, err)
+				return
+			}
+
+			itemToDelete := &model.Item{
+				ID:      itemUUID,
+				VaultID: vaultID,
+			}
+
+			// Try to delete, retry once after 1s if it fails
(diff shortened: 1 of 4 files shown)


test/e2e/utils/cleanup/cleanup.go
```diff
@@ -0,0 +1,43 @@
...
+			ID:      itemUUID,
+			VaultID: vaultID,
+		}
+
+		// Try to delete, retry once after 1s if it fails
+		err = testClient.DeleteItem(ctx, itemToDelete, vaultID)
+		if err != nil {
+			time.Sleep(1 * time.Second)
+			_ = testClient.DeleteItem(ctx, itemToDelete, vaultID)
+		}
```
JillRegan: Added retry with a delay as there is still a small chance of a 500 response from DELETE again on conflict

volodymyrZotov: `[nit]`

Does it make sense to return an error if it still fails after retry and not silence them?

test/e2e/utils/cleanup/cleanup.go
```diff
@@ -0,0 +1,43 @@
...
+	"time"
+
+	"github.com/1Password/terraform-provider-onepassword/v2/internal/onepassword/model"
+	"github.com/1Password/terraform-provider-onepassword/v2/test/e2e/utils/client"
+)
+
+// RegisterItemCleanup registers a cleanup function that will delete an item even if the test fails.
+func RegisterItemCleanup(t *testing.T, itemUUID, vaultID string) {
+	if itemUUID == "" {
+		return
```
volodymyrZotov: Do we expect `itemUUID` will be always passed? Maybe return an error in case if not?

test/e2e/utils/uuid/uuid.go
```diff
@@ -49,3 +51,17 @@ func VerifyItemUUIDUnchanged(t *testing.T, resourceName string, expectedUUID *st
 		return nil
 	}
 }
+
+// CaptureItemUUIDAndRegisterCleanup captures the UUID and registers cleanup
+func CaptureItemUUIDAndRegisterCleanup(t *testing.T, resourceName string, uuidPtr *string, vaultID string) resource.TestCheckFunc {
```
volodymyrZotov: Feels like function with such a name should not belong to `uuid` package as it does cleanup side effect.

I'd rename it or move to other package.

Maybe use something like `RegisterItem` or `WatchItem`?

test/e2e/utils/cleanup/cleanup.go
JillRegan: Updated to return an error 👍

test/e2e/utils/uuid/uuid.go
JillRegan: That makes sense. I've updated to keep the UUID capture seperate like before, and then call `RegisterItem` with the captured UUID to ensure cleanup happens.

test/e2e/utils/cleanup/cleanup.go
JillRegan: Unfortunately we can't return an error with terraform plugins Cleanup method. t.Cleanup expects a function with no return value. If this Cleanup helper runs it will because the test has already failed so that will already add some indication of an issue but I have added a log in the case that the cleanup fails.

volodymyrZotov (approved):
Looks great! 🤩

PR #295, opened 2025-12-15
Add retry on conflict helper

JillRegan:
### ✨ Summary

- This PR adds a retry helper to Create/Update/Delete methods in the case of a conflict (vault version issue). It also refactors to logic for wait which handles connect 404s to utilize the same logic.
- It also refactors the wait for eventual consistency to use helper logic
- Note -> unfortunately we can still have the errors in the case of Deletes with Connect. Connect Delete returns a generic 500 something went wrong error, even though the Connect logs indicate it was a 409 Conflict. 

### 🔗 Resolves:

Resolves: https://github.com/1Password/terraform-provider-onepassword/issues/222

Please note: This does **NOT** resolve the issue fully for DELETE method with Connect. Connect SDK does not return specific error message for the DELETE request, but rather always a generic `500 something went wrong`. 

An issue has been logged internally to address this. Once that has been updated Connect 409 error retries for DELETE will be handled here. 

### ✅ Checklist

- [x] 🖊️ Commits are signed
  - [ ] 🔹 Unit /🔸 Integration
  - [ ] 🌐 E2E

### DIFF
--- a/internal/onepassword/connect/connect_client.go
+++ b/internal/onepassword/connect/connect_client.go
@@ -136,11 +136,7 @@ func (c *Client) CreateItem(ctx context.Context, item *model.Item, vaultUuid str
 	_ = util.Retry404UntilCondition(ctx, func() (bool, error) {
 		fetchedItem, err := c.connectClient.GetItemByUUID(createdItem.ID, vaultUuid)
 		if err != nil {
-			// If error is 404, item not available yet, continue retrying
-			if strings.Contains(err.Error(), "404") || strings.Contains(err.Error(), "not found") {
-				return false, nil
-			}
-			// Other errors are not retryable
+			// 404 will be retried, others returned immediately
 			return false, err
 		}
 		// Item exists, check if it has version 1 (newly created)


volodymyrZotov:
> Connect Delete returns a generic 500 something went wrong error, even though the Connect logs indicate it was a 409 Conflict.

We'll need to think about the fix for that and address in future PR. Or maybe on Connect side.

internal/onepassword/util/retry.go
```diff
@@ -0,0 +1,31 @@
...
+		if err == nil {
+			// Operation succeeded
+			return nil
+		}
+
+		errStr := err.Error()
+		if strings.Contains(errStr, "409") || strings.Contains(errStr, "Conflict") || strings.Contains(errStr, "conflict") {
+			// 409 error - retry
+			continue
+		}
```
volodymyrZotov: The implementation is good, but I was thinking whether we can make it more robust.

As we know that conflict happens due to the vault version mismatch. Can we rely on it as well? So the flow will look like:

1. There is 409 error
2. It makes get item request to get the latest vault version
3. Set latest vault version from fetched item and retries the operation
4. Retry again in case of an error.

Also in addition to that we can use [Exponential Backoff with Jitter](https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/) algorithm similar to what we have in [wait](https://github.com/1Password/terraform-provider-onepassword/blob/main/internal/onepassword/connect/connect_client.go#L289) function.

In addition keeping in mid that we already have retries implemented in Connect client to handle 404 errors, it worth to think whether we can/want to refactor it to align with this new RetryOnConflict function.
Maybe at least extract common code to the `util`.

How does it sound?

internal/onepassword/util/retry.go
```diff
@@ -0,0 +1,31 @@
...
+
+import (
+	"context"
+	"fmt"
+	"strings"
+)
+
+// RetryOnConflict retries an operation when it returns 409 Conflict errors
+func RetryOnConflict(ctx context.Context, operation func() error) error {
+	maxAttempts := 3
```
volodymyrZotov: Can we make this configurable? So, users can configure it by themselves.

internal/provider/provider.go
```diff
@@ -92,6 +93,10 @@ func (p *OnePasswordProvider) Schema(ctx context.Context, req provider.SchemaReq
 				Description: "A valid account name or ID to use desktop app authentication. Can also be sourced from `OP_ACCOUNT` environment variable.",
 				Optional:    true,
 			},
+			"max_retries": schema.Int64Attribute{
+				MarkdownDescription: "Maximum number of retry attempts for Connect operations (defaults to 5). Only applies when using Connect server.",
```
volodymyrZotov: This is applicable for Connect and SDK, correct?

internal/provider/provider.go
```diff
@@ -92,6 +93,10 @@ func (p *OnePasswordProvider) Schema(ctx context.Context, req provider.SchemaReq
 				Description: "A valid account name or ID to use desktop app authentication. Can also be sourced from `OP_ACCOUNT` environment variable.",
 				Optional:    true,
 			},
+			"max_retries": schema.Int64Attribute{
+				MarkdownDescription: "Maximum number of retry attempts for operations (defaults to 5).",
+				Optional:            true,
+			},
```
volodymyrZotov: Let's add validator that makes sure user provided values `>= 0`
```
Validators: []validator.Int64{
    int64validator.AtLeast(0),
},
```

internal/onepassword/connect/connect_client.go
```diff
@@ -256,47 +272,23 @@ func (c *Client) GetFileContent(_ context.Context, file *model.ItemFile, itemUUI
...
-		}
-
-		// Condition not met yet, continue retrying
-		// Exponential backoff: 50ms, 100ms, 200ms, 400ms, etc., capped at 500ms
-		backoff := time.Duration(50*(1<<uint(attempt))) * time.Millisecond
-		if backoff > 500*time.Millisecond {
-			backoff = 500 * time.Millisecond
-		}
+// refreshVersion fetches the latest item version and updates both connectItem and model item
+func (c *Client) refreshVersion(itemID, vaultUuid string, connectItem *onepassword.Item, item *model.Item) error {
```
volodymyrZotov: This retry function should make `GET /vaults/:id` to fetch latest vault and take `contentVersion`.

Then it sets it to `item.VaultVersion = vault.ContentVersion` and retry.

Something like:
```
func (c *Client) refreshVaultVersion(ctx context.Context, vaultUuid string, item *model.Item) error {
	vault, err := c.GetVault(ctx, vaultUuid)
	if err != nil {
		return fmt.Errorf("failed to fetch latest vault version: %w", err)
	}
	item.VaultVersion = vault.ContentVersion
	return nil
}
```

internal/onepassword/connect/connect_client.go
```diff
@@ -125,16 +123,21 @@ func (c *Client) CreateItem(ctx context.Context, item *model.Item, vaultUuid str
 		return nil, err
 	}
 
-	createdItem, err := c.connectClient.CreateItem(connectItem, vaultUuid)
+	var createdItem *onepassword.Item
+	err = util.RetryOnConflict(ctx, func() error {
+		var createErr error
+		createdItem, createErr = c.connectClient.CreateItem(connectItem, vaultUuid)
+		return createErr
+	}, nil, c.maxRetries())
```
volodymyrZotov: We should pass retry function for each operation (create/update/delete).

internal/onepassword/util/retry.go
```diff
@@ -0,0 +1,97 @@
...
+	return fmt.Errorf("max retry attempts (%d) reached for operation due to %s", maxAttempts, errorMsg)
+}
+
+// RetryOnConflict retries an operation when it returns 409 Conflict errors
+func RetryOnConflict(ctx context.Context, operation func() error) error {
+	return retry(ctx, operation, []string{"409", "Conflict", "conflict"}, "409 Conflict errors", maxRetryAttempts)
+}
+
+// RetryUntilCondition retries an operation when it returns 404 Not Found errors
+func RetryUntilCondition(ctx context.Context, operation func() (bool, error)) error {
```
volodymyrZotov:
```suggestion
func Retry404UntilCondition(ctx context.Context, operation func() (bool, error)) error {
```

internal/onepassword/util/retry.go
```diff
@@ -0,0 +1,97 @@
...
+		if attempt < maxAttempts-1 {
+			if ctx.Err() != nil {
+				return ctx.Err()
+			}
+			backoff := calculateBackoff(attempt)
+			time.Sleep(backoff)
+		}
+	}
+
+	return fmt.Errorf("max retry attempts (%d) reached for operation due to %s", maxAttempts, errorMsg)
```
volodymyrZotov: Why do we need to return `errorMsg` instead of the error that happened on line 19 `err := operation()`?

Current implementation overwrites original error.

internal/onepassword/connect/connect_client.go
volodymyrZotov: As discussed on a call, SDKs doesn't support vault version on the Item entity, therefore it's not possible to pass it within request. So, we simply retry without relying on vault version.

internal/provider/provider.go
volodymyrZotov: No relevant anymore.

internal/onepassword/util/retry.go
```diff
@@ -0,0 +1,97 @@
...
+
+	exponentialDelay := baseDelay * time.Duration(1<<uint(attempt))
+
+	if exponentialDelay > maxDelay {
+		exponentialDelay = maxDelay
+	}
+
+	jitter := time.Duration(rand.Int63n(int64(baseDelay)))
+
+	return exponentialDelay + jitter
```
volodymyrZotov: This will return the number bigger than `maxDelay`.

Seems it should be
```
delay := exponentialDelay + jitter
if delay > maxDelay {
    delay = maxDelay
}
return delay
```

volodymyrZotov (approved):

PR #299, opened 2025-12-16
Refactor itemToData

JillRegan:
### ✨ Summary

This PR moves logic for building an item to helper methods in order to improve readability. It also adds test cases to the builders.

### 🔗 Resolves:

- This does not resolve a specific issue, rather it just improves readability.

### ✅ Checklist

- [x] 🖊️ Commits are signed
  - [x] 🔹 Unit /🔸 Integration
  - [ ] 🌐 E2E

### DIFF
--- a/internal/provider/model_to_state_test.go
+++ b/internal/provider/model_to_state_test.go
@@ -16,8 +16,8 @@ func TestToStateTags(t *testing.T) {
 	testCases := []struct {
 		name          string
 		modelTags     []string
-		stateTags     []string // Use []string for easier initialization
-		wantTags      []string // Use []string for easier initialization
+		stateTags     []string
+		wantTags      []string
 		stateTagsNull bool
 		wantNull      bool
 		wantErr       bool
@@ -68,7 +68,7 @@ func TestToStateTags(t *testing.T) {
 			name:      "tags with special characters",
 			modelTags: []string{"tag-1", "tag_2", "tag.3", "tag@4"},
 			stateTags: []string{"tag-1", "tag_2", "tag.3", "tag@4"},
-			wantTags:  []string{"tag-1", "tag_2", "tag.3", "tag@4"}, // Correct lexicographic order
+			wantTags:  []string{"tag-1", "tag_2", "tag.3", "tag@4"},
 		},
 		{
 			name:      "empty string in tags",


internal/provider/onepassword_item_resource.go
volodymyrZotov: What do you think about renaming this function to `modelToState` so it better reflects what it does?

Also what do you think about renaming `data` parameter to `state` and even maybe `item` to `model`?

If we ok with that that will also affect function in item_to_data.go and file name as well.

internal/provider/model_to_state.go
```diff
@@ -0,0 +1,144 @@
...
+)
+
+func processTags(ctx context.Context, itemTags []string, currentTags types.List) (types.List, diag.Diagnostics) {
+	var dataTagsSlice []string
+	diagnostics := currentTags.ElementsAs(ctx, &dataTagsSlice, false)
+	if diagnostics.HasError() {
+		return currentTags, diagnostics
+	}
+
+	sort.Strings(dataTagsSlice)
```
volodymyrZotov, line 21: I think we also want to sort item tags as well `sort.Strings(itemTags)` to make sure the order is the same.

Would be good to have unit test for this case

internal/provider/item_to_data_test.go
```diff
@@ -0,0 +1,322 @@
+package provider
```
volodymyrZotov: Could you please take a look into the following scenarios and check if we need to cover them with additional tests:

*processTags*:
- Item tags in different order should still match (e.g., ["b", "a"] vs ["a", "b"])

*processSectionsAndFields*:
- Match section by label when ID is empty
- Match field by label when ID is empty
- Update existing field (not just create new)
- Multiple sections with multiple fields
- Field with SectionID that doesn't match any section (should be ignored)

*processTopLevelFields*:
- "server" label should map to Hostname (currently only "hostname" is tested)
- if `f.Purpose` is empty "username" should be set by label.
- if `f.Purpose` is empty "password" should be set by label.
- Verify existing data values are preserved when field not present in item (In Terraform, when you do a Read operation:
State has existing values (e.g., password = "abc123")
1Password item might not return certain fields (e.g., API doesn't include empty fields)
If the function accidentally clears fields not present in the response, you get state drift
The current implementation is correct (only sets values when field is present in the loop), but there's no test verifying this behavior.)

internal/provider/model_to_state.go
JillRegan: Good call out 👍  

I have added handdling to sort the itemTags (now modelTags)

internal/provider/item_to_data_test.go
JillRegan: Added the following tests:

`processTags`:

- Item tags in different order should still match (e.g., ["b", "a"] vs ["a", "b"]) ✅ 

I also added tests for
- Single tag
- Tags with special characters
- Empty string in tags
- State has more tags


`processSectionsAndFields`:

- Match section by label when ID is empty ✅ 
- Match field by label when ID is empty ✅ 
- Update existing field (not just create new) ✅ 
- Multiple sections with multiple fields ✅ 
- Field with SectionID that doesn't match any section (should be ignored) ✅ 

Also added tests for:
- Section with empty label
- Field with empty label
- Recipe with no charset


`processTopLevelFields`:

- "server" label should map to Hostname (currently only "hostname" is tested) ✅ 
- if f.Purpose is empty "username" should be set by label. ✅ 
- if f.Purpose is empty "password" should be set by label. ✅ 
- Verify existing data values are preserved when field not present in item ✅ 

Also added:
- Null state preserved when field has not been set
- Unknown state preserved when field has not been set

internal/provider/onepassword_item_resource.go
JillRegan: I think that is a great idea! I have updated the naming here.

Only difference to your suggestion is instead of changing `item` to `model`, I updated it to `itemModel`. Because we already import model as a package it would cause conflicts with what is being referenced:

```
import (
	"github.com/1Password/terraform-provider-onepassword/v2/internal/onepassword/model"
)
```

internal/provider/model_to_state_test.go
```diff
@@ -0,0 +1,741 @@
...
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			// Set up test-specific values
+			var stateTags, want types.List
+			switch tt.name {
+			case "tags match":
+				stateTags, _ = types.ListValueFrom(ctx, types.StringType, []string{"tag1", "tag2"})
+				want, _ = types.ListValueFrom(ctx, types.StringType, []string{"tag1", "tag2"})
```
volodymyrZotov: Instead of having such a long switch case, I'd suggest to add additional properties to the test struct `stateTags` and `want`, so you define all parameters there, and can completely remove switch/case.

volodymyrZotov (approved):
I like this refactoring so much ❤️. It's so easy to read it now.
Great job! 🌟 

Put a small non blocking comment about how tests readability can be improved. Approved!

PR #303, opened 2025-12-31
Build section map for data source

JillRegan:
### ✨ Summary

- This PR adds a section map to data source to make it easier to access specific values. It also adds unit/e2e test.

### 🔗 Resolves:

N/A

### ✅ Checklist

- [x] 🖊️ Commits are signed
  - [x] 🔹 Unit /🔸 Integration
  - [x] 🌐 E2E

- Targets feature branch

### DIFF
--- a/docs/data-sources/item.md
+++ b/docs/data-sources/item.md
@@ -57,7 +57,7 @@ data "onepassword_item" "example" {
 
 Optional:
 
-- `field_map` (Attributes Map) A map of customfields in the section, keyed by field label. (see [below for nested schema](#nestedatt--section_map--field_map))
+- `field_map` (Attributes Map) A map of custom fields in the section, keyed by field label. (see [below for nested schema](#nestedatt--section_map--field_map))
 - `file_map` (Attributes Map) A map of files attached to the section, keyed by file ID. (see [below for nested schema](#nestedatt--section_map--file_map))
 
 Read-Only:
--- a/internal/provider/const.go
+++ b/internal/provider/const.go
@@ -52,7 +52,7 @@ const (
 	fieldPurposeDescription = "Purpose indicates this is a special field: a username, password, or notes field."
 	fieldTypeDescription    = "The type of value stored in the field."
 	fieldValueDescription   = "The value of the field."
-	fieldMapDescription     = "A map of customfields in the section, keyed by field label."
+	fieldMapDescription     = "A map of custom fields in the section, keyed by field label."
 
 	passwordRecipeDescription  = "The recipe used to generate a new value for a password."
 	passwordElementDescription = "The kinds of characters to include in the password."
--- a/internal/provider/onepassword_item_data_source_test.go
+++ b/internal/provider/onepassword_item_data_source_test.go
@@ -280,8 +280,79 @@ func TestAccItemDataSourceSectionMap(t *testing.T) {
 					resource.TestCheckResourceAttr("data.onepassword_item.test", "uuid", expectedItem.ID),
 					resource.TestCheckResourceAttr("data.onepassword_item.test", "category", strings.ToLower(string(expectedItem.Category))),
 					resource.TestCheckResourceAttr("data.onepassword_item.test", fmt.Sprintf("section_map.%s.id", expectedItem.Sections[0].Label), expectedItem.Sections[0].ID),
+				),
+			},
+		},
+	})
+}
+
+func TestAccItemDataSourceSectionMapFieldMap(t *testing.T) {
+	expectedItem := generateItemWithSections()
+	expectedVault := model.Vault{
+		ID:          expectedItem.VaultID,
+		Name:        "Name of the vault",
+		Description: "This vault will be retrieved",
+	}
+
+	testServer := setupTestServer(expectedItem, expectedVault, t)
+	defer testServer.Close()
+
+	resource.Test(t, resource.TestCase{
+		ProtoV6ProviderFactories: testAccProtoV6ProviderFactories,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccProviderConfig(testServer.URL) + testAccItemDataSourceConfig(expectedItem.VaultID, expectedItem.ID),
+				Check: resource.ComposeAggregateTestCheckFunc(
+					resource.TestCheckResourceAttr("data.onepassword_item.test", "id", fmt.Sprintf("vaults/%s/items/%s", expectedVault.ID, expectedItem.ID)),
+					resource.TestCheckResourceAttr("data.onepassword_item.test", "vault", expectedVault.ID),
+					resource.TestCheckResourceAttr("data.onepassword_item.test", "title", expectedItem.Title),
+					resource.TestCheckResourceAttr("data.onepassword_item.test", "uuid", expectedItem.ID),
+					resource.TestCheckResourceAttr("data.onepassword_item.test", "category", strings.ToLower(string(expectedItem.Category))),
+					resource.TestCheckResourceAttr("data.onepassword_item.test", fmt.Sprintf("section_map.%s.field_map.%s.id", expectedItem.Sections[0].Label, expectedItem.Fields[0].Label), expectedItem.Fields[0].ID),
 					resource.TestCheckResourceAttr("data.onepassword_item.test", fmt.Sprintf("section_map.%s.field_map.%s.value", expectedItem.Sections[0].Label, expectedItem.Fields[0].Label), expectedItem.Fields[0].Value),
 					resource.TestCheckResourceAttr("data.onepassword_item.test", fmt.Sprintf("section_map.%s.field_map.%s.type", expectedItem.Sections[0].Label, expectedItem.Fields[0].Label), string(expectedItem.Fields[0].Type)),
+				),
+			},
+		},
+	})
+}
+
+func TestAccItemDataSourceSectionMapFileMap(t *testing.T) {
+	expectedItem := generateLoginItemWithFiles()
+	expectedVault := model.Vault{
+		ID:          expectedItem.VaultID,
+		Name:        "Name of the vault",
+		Description: "This vault will be retrieved",
+	}
+
+	testServer := setupTestServer(expectedItem, expectedVault, t)
+	defer testServer.Close()
+
+	first_content, err := expectedItem.Files[0].Content()
+	if err != nil {
+		t.Fatalf("Error getting content of first file: %v", err)
+	}
+
+	second_content, err := expectedItem.Files[1].Content()
+	if err != nil {
+		t.Fatalf("Error getting content of second file: %v", err)
+	}
+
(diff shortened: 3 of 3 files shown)


test/e2e/item_resource_test.go
```diff
@@ -1229,6 +1229,7 @@ func TestAccItemResourcePasswordWriteOnly(t *testing.T) {
 				Check: resource.ComposeAggregateTestCheckFunc(
 					logStep(t, "CREATE_WITH_PASSWORD_WO"),
 					uuidutil.CaptureItemUUID(t, "onepassword_item.test_item", &itemUUID),
+					cleanup.RegisterItem(t, &itemUUID, testVaultID),
```
JillRegan, line 1232: Saw this was missing the cleanup helper so just added in this PR.

docs/data-sources/item.md
```diff
@@ -46,6 +46,7 @@ data "onepassword_item" "example" {
 - `private_key_openssh` (String, Sensitive) SSH Private key in OpenSSH format.
 - `public_key` (String) SSH Public Key for this item.
 - `section` (Block List) A list of custom sections in an item (see [below for nested schema](#nestedblock--section))
+- `section_map` (Attributes Map) A map of sections in the item, keyed by section label. This allows easy lookup of sections and fields by name. (see [below for nested schema](#nestedatt--section_map))
```
volodymyrZotov: too keep it consistent with `section` (list)
```suggestion
- `section_map` (Attributes Map) A map of custom sections in the item, keyed by section label. This allows easy lookup of sections and fields by name. (see [below for nested schema](#nestedatt--section_map))
```

internal/provider/onepassword_item_data_source.go
```diff
@@ -53,6 +55,7 @@ type OnePasswordItemDataSourceModel struct {
 	PrivateKey        types.String                  `tfsdk:"private_key"`
 	PrivateKeyOpenSSH types.String                  `tfsdk:"private_key_openssh"`
 	Section           []OnePasswordItemSectionModel `tfsdk:"section"`
```
volodymyrZotov: What do you think about renaming this property to `SectionList` so it clearly indicates what data structure is behind it? Similar I've done for [the resource](https://github.com/1Password/terraform-provider-onepassword/pull/309/changes#diff-00f1176358c6cdc797c4af55190b4e5ee2fa897b9ada46e89388a150dc50214cR61).

internal/provider/onepassword_item_data_source.go
```diff
@@ -53,6 +55,7 @@ type OnePasswordItemDataSourceModel struct {
...
+	SectionMap        types.Map                     `tfsdk:"section_map"`
```
volodymyrZotov: Maybe change the type to be `map[string]OnePasswordItemSectionMapModel` to be more explicit?

internal/provider/onepassword_item_data_source.go
```diff
@@ -78,6 +81,19 @@ type OnePasswordItemFieldModel struct {
 	Value   types.String `tfsdk:"value"`
 }
 
+type OnePasswordItemSectionMapModel struct {
+	ID       types.String                                   `tfsdk:"id"`
+	FieldMap map[string]OnePasswordItemSectionMapFieldModel `tfsdk:"field_map"`
+	File     []OnePasswordItemFileModel                     `tfsdk:"file"`
```
volodymyrZotov: Should we change data structure from list to map here? So files can also be discoverable by label?

internal/provider/onepassword_item_data_source.go
```diff
@@ -198,7 +214,71 @@ func (d *OnePasswordItemDataSource) Schema(ctx context.Context, req datasource.S
...
+			"section_map": schema.MapNestedAttribute{
+				MarkdownDescription: "A map of sections in the item, keyed by section label. This allows easy lookup of sections and fields by name.",
+				Computed:            true,
+				NestedObject: schema.NestedAttributeObject{
+					Attributes: map[string]schema.Attribute{
+						"id": schema.StringAttribute{
+							MarkdownDescription: sectionIDDescription,
+							Computed:            true,
+						},
+						"field_map": schema.MapNestedAttribute{
```
volodymyrZotov, line 230: Seems this should be optional as well.

In the list `schema.ListNestedBlock` is optional by default, so we can skip it. But for maps we should set this property explicitly.

internal/provider/onepassword_item_data_source.go
```diff
@@ -198,7 +214,71 @@ func (d *OnePasswordItemDataSource) Schema(ctx context.Context, req datasource.S
...
+							MarkdownDescription: "A map of fields in the section, keyed by field label.",
+							Computed:            true,
+							NestedObject: schema.NestedAttributeObject{
+								Attributes: map[string]schema.Attribute{
+									"id": schema.StringAttribute{
+										MarkdownDescription: fieldIDDescription,
+										Computed:            true,
+									},
+									"label": schema.StringAttribute{
```
volodymyrZotov: Do we want to keep label here? As label is the map's key.

internal/provider/onepassword_item_data_source.go
```diff
@@ -198,7 +214,71 @@ func (d *OnePasswordItemDataSource) Schema(ctx context.Context, req datasource.S
...
+									},
+									"value": schema.StringAttribute{
+										MarkdownDescription: fieldValueDescription,
+										Computed:            true,
+										Sensitive:           true,
+									},
+								},
+							},
+						},
+						"file": schema.ListNestedAttribute{
```
volodymyrZotov: Do we want to change file data structure to map as well? So files can be discoverable by label similar to the fields?

internal/provider/onepassword_item_data_source.go
```diff
@@ -198,7 +214,71 @@ func (d *OnePasswordItemDataSource) Schema(ctx context.Context, req datasource.S
...
+							MarkdownDescription: sectionFilesDescription,
+							Computed:            true,
+							NestedObject: schema.NestedAttributeObject{
+								Attributes: map[string]schema.Attribute{
+									"id": schema.StringAttribute{
+										MarkdownDescription: fileIDDescription,
+										Computed:            true,
+									},
+									"name": schema.StringAttribute{
```
volodymyrZotov: If we change `file` datastructure to `map`, name can be removed as `name` will be the map's key.

internal/provider/onepassword_item_data_source.go
```diff
@@ -198,7 +214,71 @@ func (d *OnePasswordItemDataSource) Schema(ctx context.Context, req datasource.S
 				Computed:            true,
 				Sensitive:           true,
 			},
+			"section_map": schema.MapNestedAttribute{
```
volodymyrZotov, line 220: Seems this should be optional.

In the list `schema.ListNestedBlock` is optional by default, so we can skip it. But for maps we should set this property explicitly.

internal/provider/onepassword_item_data_source.go
```diff
@@ -437,3 +524,88 @@ func getItemForDataSource(ctx context.Context, client onepassword.Client, data O
...
+func buildSectionMap(ctx context.Context, item *model.Item, client onepassword.Client) (types.Map, diag.Diagnostics) {
+	var diagnostics diag.Diagnostics
+
+	sectionMap := make(map[string]OnePasswordItemSectionMapModel)
+
+	for _, s := range item.Sections {
+		sectionLabel := s.Label
+		if sectionLabel == "" {
+			// Use the section ID as the label if the section has no label
+			sectionLabel = s.ID
```
volodymyrZotov: As per our conversation we allow sections with empty labels. So do not need to use ID as a label

internal/provider/onepassword_item_data_source.go
```diff
@@ -437,3 +524,88 @@ func getItemForDataSource(ctx context.Context, client onepassword.Client, data O
...
+		}
+
+		fieldMap := make(map[string]OnePasswordItemSectionMapFieldModel)
+
+		for _, f := range item.Fields {
+			if f.SectionID != "" && f.SectionID == s.ID {
+				fieldLabel := f.Label
+				if fieldLabel != "" {
```
volodymyrZotov: As per the conversation we had, we allow fields with empty labels.

internal/provider/const.go
```diff
@@ -36,20 +36,23 @@ const (
...
 	fileContentBase64Description = "The content of the file in base64 encoding. (Use this for binary files.)"
+	fileMapDescription           = "A map of files attached to the section, keyed by file ID."
 
 	fieldDescription        = "A custom field."
 	fieldIDDescription      = "A unique identifier for the field."
 	fieldLabelDescription   = "The label for the field."
 	fieldPurposeDescription = "Purpose indicates this is a special field: a username, password, or notes field."
 	fieldTypeDescription    = "The type of value stored in the field."
 	fieldValueDescription   = "The value of the field."
+	fieldMapDescription     = "A map of customfields in the section, keyed by field label."
```
volodymyrZotov:
```suggestion
	fieldMapDescription     = "A map of custom fields in the section, keyed by field label."
```

internal/provider/onepassword_item_data_source_test.go
```diff
@@ -257,6 +257,37 @@ func TestAccItemSSHKey(t *testing.T) {
...
+				Check: resource.ComposeAggregateTestCheckFunc(
+					resource.TestCheckResourceAttr("data.onepassword_item.test", "id", fmt.Sprintf("vaults/%s/items/%s", expectedVault.ID, expectedItem.ID)),
+					resource.TestCheckResourceAttr("data.onepassword_item.test", "vault", expectedVault.ID),
+					resource.TestCheckResourceAttr("data.onepassword_item.test", "title", expectedItem.Title),
+					resource.TestCheckResourceAttr("data.onepassword_item.test", "uuid", expectedItem.ID),
+					resource.TestCheckResourceAttr("data.onepassword_item.test", "category", strings.ToLower(string(expectedItem.Category))),
+					resource.TestCheckResourceAttr("data.onepassword_item.test", fmt.Sprintf("section_map.%s.id", expectedItem.Sections[0].Label), expectedItem.Sections[0].ID),
+					resource.TestCheckResourceAttr("data.onepassword_item.test", fmt.Sprintf("section_map.%s.field_map.%s.value", expectedItem.Sections[0].Label, expectedItem.Fields[0].Label), expectedItem.Fields[0].Value),
+					resource.TestCheckResourceAttr("data.onepassword_item.test", fmt.Sprintf("section_map.%s.field_map.%s.type", expectedItem.Sections[0].Label, expectedItem.Fields[0].Label), string(expectedItem.Fields[0].Type)),
```
volodymyrZotov, line 313: Could you also check attached files as well?

JillRegan: Ended up separating out the test for each map.

volodymyrZotov (approved):

Issue #212, opened 2025-02-23
Support for Ephemeral Values in 1Password Terraform Provider to Enhance Secret Security


aurimasniekis:
### Summary  
The **1Password Terraform provider** should support **Terraform 1.10's ephemeral values** to ensure that secrets retrieved from 1Password are **never persisted in Terraform state or plan files**. This is a critical improvement for organizations that need to enforce strict security policies while managing secrets in their infrastructure.

#### **Secure Infrastructure Provisioning**  
Organizations provisioning infrastructure often require secrets from 1Password, such as database credentials, API keys, or certificates. Without ephemeral values, these secrets risk being stored in Terraform state files, increasing the potential for exposure through misconfiguration, state file compromise, or unauthorized access.  

#### **Compliance Requirements**  
Many enterprises operate under strict security frameworks such as **SOC 2, PCI-DSS, ISO 27001, and NIST**, which mandate that sensitive information like credentials must not be stored in plaintext or retained longer than necessary. Even encrypted Terraform state files may not meet compliance requirements if secrets can still be extracted.  

#### **GitOps and CI/CD Security**  
Teams following **GitOps** principles or using **CI/CD pipelines** need to ensure that secrets retrieved from 1Password are not **committed to version control** or exposed in Terraform state artifacts. Many organizations rely on remote state backends, such as Terraform Cloud or AWS S3, where improperly handled secrets could be accessible to a broader audience than intended.  

#### **Ephemeral and Short-Lived Credentials**  
Some workflows rely on **dynamic secret generation and rotation**, where credentials are meant to be short-lived. If secrets are stored in Terraform state, old credentials can be unintentionally restored, leading to operational failures or security vulnerabilities. Ephemeral values would ensure that secrets are only used within the current Terraform execution cycle and are **never persisted** beyond that.  

#### **Least Privilege and Access Control**  
By leveraging ephemeral values, **role-based access control (RBAC) policies** can be better enforced. Infrastructure teams can limit who has access to secrets while ensuring that Terraform itself does not retain sensitive data in a way that circumvents security policies.  

### Proposed solution  
The **1Password Terraform provider** should integrate **ephemeral resources** and **ephemeral input/output variables** to align with Terraform 1.10’s security model. This will allow secrets to be **retrieved at runtime, used for configuration, and discarded immediately**—never persisting in any Terraform-generated artifacts.  

### Is there a workaround to accomplish this today?  
Currently, organizations must resort to **manual workarounds** such as:  
[...]

PR #304, opened 2026-01-02
Add write only note value 

JillRegan:
### ✨ Summary

- This PR adds a write only note value field (`note_value_wo` and `note_value_version_wo`) to the resource item in order to allow Ephemeral values for notes. 
- Matching with write only password implementation [here](https://github.com/1Password/terraform-provider-onepassword/pull/229) 
- Refactors some logic for `password_wo` to use helpers 
- Like `password_wo`, `note_value_wo `needs to have `note_value_version_wo` set in order to be applied. To update the version must be increased. 

### 🔗 Resolves:

https://github.com/1Password/terraform-provider-onepassword/issues/212 (Does not full resolve as we need to add more write only fields for other sensitive values)

### ✅ Checklist

- [x] 🖊️ Commits are signed
  - [x] 🔹 Unit /🔸 Integration
  - [x] 🌐 E2E

To manually tests no regressions to note_value test with this config:

```
resource "onepassword_item" "test_note_value" {
  vault    = data.onepassword_vault.terraform.uuid
  title    = "Test Note Value"
  category = "secure_note"

  note_value = <<EOT
This is a regular note value.
It can be read and written.
Multi-line support works here.
EOT
}
```

To test creating a note_value_write_only use this config:

```
resource "onepassword_item" "test_note_value_wo" {
  vault    = data.onepassword_vault.terraform.uuid
  title    = "Test Note Value Write-Only"
  category = "secure_note"

  note_value_wo         = <<EOT
This is a write-only note value.
It will not be stored in state.
Version 1
EOT
  note_value_wo_version = 1
}
```

To test note_value can be updated use: 

```
resource "onepassword_item" "test_note_value_wo" {
  vault    = data.onepassword_vault.terraform.uuid
  title    = "Test Note Value Write-Only"
  category = "secure_note"

  note_value_wo         = <<EOT
This is a write-only note value.
It will not be stored in state.
Version 2 Update
EOT
  note_value_wo_version = 2
}
```

To ensure user cannot set note_value and note_value_wo use: 

(will produce error - expected)

```
resource "onepassword_item" "test_note_value_wo_update" {
   vault    = data.onepassword_vault.terraform.uuid
   title    = "Test Note Value Write-Only"
   category = "secure_note"
   note_value = "test"

   note_value_wo         = <<EOT
 This is a write-only note value.
 It will not be stored in state.
 Version 3 - Updated!
 EOT
   note_value_wo_version = 3
 }
```

### DIFF
--- a/internal/provider/ephemeral_values.go
+++ b/internal/provider/ephemeral_values.go
@@ -28,22 +28,9 @@ func handleWriteOnlyFieldUpdates(
 	plan *OnePasswordItemResourceModel,
 	refreshItem func() (*model.Item, error),
 ) error {
-	// Helper to check if a field needs the current item
-	needsItem := func(configVersion, stateVersion types.Int64) bool {
-		if configVersion.IsNull() {
-			return false
-		}
-		configVer := configVersion.ValueInt64()
-		stateVer := int64(0)
-		if !stateVersion.IsNull() {
-			stateVer = stateVersion.ValueInt64()
-		}
-		return configVer <= stateVer
-	}
-
 	// Check if any write-only field needs the current item
-	passwordNeedsItem := needsItem(config.PasswordWOVersion, state.PasswordWOVersion)
-	noteValueNeedsItem := needsItem(config.NoteValueWOVersion, state.NoteValueWOVersion)
+	passwordNeedsItem := shouldFetchCurrentItem(config.PasswordWOVersion, state.PasswordWOVersion)
+	noteValueNeedsItem := shouldFetchCurrentItem(config.NoteValueWOVersion, state.NoteValueWOVersion)
 
 	// Fetch item once if needed
 	var currentItem *model.Item
@@ -106,4 +93,17 @@ func handleWriteOnlyFieldUpdates(
 	}
 
 	return nil
+}
+
+// shouldFetchCurrentItem returns true if the write-only field version hasn't increased
+func shouldFetchCurrentItem(configVersion, stateVersion types.Int64) bool {
+	if configVersion.IsNull() {
+		return false
+	}
+	configVer := configVersion.ValueInt64()
+	stateVer := int64(0)
+	if !stateVersion.IsNull() {
+		stateVer = stateVersion.ValueInt64()
+	}
+	return configVer <= stateVer
 }


internal/provider/ephemeral_values.go
```diff
@@ -0,0 +1,72 @@
...
+		stateVer = stateVersion.ValueInt64()
+	}
+
+	if configVer > stateVer {
+		// Version increased - use new write-only value
+		*planValue = woValue
+	} else {
+		// Version unchanged or decreased - preserve existing value by reading current item
+		vaultUUID, itemUUID := vaultAndItemUUID(planID.ValueString())
+		currentItem, err := r.client.GetItem(ctx, itemUUID, vaultUUID)
```
volodymyrZotov: With current implementation it will get an item as many times as write only fields is used. Right now it will make get request 2 times, once for `password` and once `secure_note`.

I'd like us to optimize this and fetch an item only once and use it to update all required write only values.

internal/provider/ephemeral_values.go
```diff
@@ -0,0 +1,72 @@
...
+}
+
+// clearWriteOnlyFieldFromState clears a field from state if write-only version is set
+func clearWriteOnlyFieldFromState(version types.Int64, stateValue *types.String) {
+	if !version.IsNull() {
+		*stateValue = types.StringNull()
+	}
+}
+
+func (r *OnePasswordItemResource) handleWriteOnlyFieldUpdate(
```
volodymyrZotov: I see the idea of moving this method into separate file to improve readability. But as this is method of `OnePasswordItemResource`, I'd expect to see all the method definitions in the same file where the struct is defined.

I'd suggest to keep this function here as helper but not make it a method. And instead of directly using `r.client.GetItem` pass it as parameter `refreshItem func() *model.Item`.

internal/provider/ephemeral_values.go
```diff
@@ -0,0 +1,72 @@
...
+	ctx context.Context,
+	configVersion types.Int64,
+	stateVersion types.Int64,
+	woValue types.String,
+	planValue *types.String,
+	planID types.String,
+	fieldPurpose model.ItemFieldPurpose,
+	fieldName string,
```
volodymyrZotov: Could we reduce an amount of parameters here? Having `8` parameters makes feeling that this function is overcomplicated.

One of possible ways could be to create additional struct or several structs that keep the parameters required for this function as struct properties.

internal/provider/ephemeral_values.go
JillRegan: Good call out. I've updated to just make this request once when needed.

internal/provider/ephemeral_values.go
JillRegan: This makes sense, I have converted to function and massing the GetItem request to the helper as a param. 👍

internal/provider/ephemeral_values.go
JillRegan: Refactored this function to take in the plan/config/state which means these params are no longer needed.

internal/provider/ephemeral_values.go
```diff
@@ -0,0 +1,109 @@
...
+func shouldFetchCurrentItem(configVersion, stateVersion types.Int64) bool {
+	if configVersion.IsNull() {
+		return false
+	}
+	configVer := configVersion.ValueInt64()
+	stateVer := int64(0)
+	if !stateVersion.IsNull() {
+		stateVer = stateVersion.ValueInt64()
+	}
+	return configVer <= stateVer
```
volodymyrZotov, line 108: This looks incorrect, as it should fetch an item only when config version > than the one in state.

JillRegan: If `configVer > stateVer` we should be using the WO value set in the config, but if it is not we need to fetch the item to use the current value instead of the WO value.

So for version increases write the new WO value from the config. When it doesn't we need to fetch the item and preserve current value.

volodymyrZotov: Ah, you are right. 👍

Issue #306, opened 2026-01-04
failed to get item using sdk: invalid user │ input: encountered the following errors: the provided ID is not a valid │ UUID


dbrennand:
### Your environment

Terraform Provider Version: 3.0.1

- Service Account: Using service account token

Authentication Method: Service Account

OS: MacOS

Terraform Version: 

```
❯ tofu version  
OpenTofu v1.11.1
on darwin_arm64
+ provider registry.opentofu.org/1password/onepassword v3.0.1
+ provider registry.opentofu.org/bpg/proxmox v0.89.1
+ provider registry.opentofu.org/hashicorp/http v3.5.0
+ provider registry.opentofu.org/hetznercloud/hcloud v1.57.0
```

## What happened?

```
# .env
OP_SERVICE_ACCOUNT_TOKEN=op://Home-Ops/soyelbgbxvtfeiwch7vkbgrorq/credential
```

```
# op.tf
provider "onepassword" {
}
```

```
op run --env-file=./.env -- tofu plan
│ Unable to read item, got error: failed to get item using sdk: invalid user
│ input: encountered the following errors: the provided ID is not a valid
│ UUID
```

## What did you expect to happen?

The provider seems to be using the SDK method instead of the Service Account as exported via the documented environment variable:

<img width="604" height="89" alt="Image" src="https://github.com/user-attachments/assets/418a42ea-2cfb-4dfd-82e9-9c6c0356d4f4" />

JillRegan:
Hi @dbrennand 👋 

The Service Account Token is the authentication method for the SDK. Previously the Service Account Token was used to authenticate the CLI with the provider, but with [v3.0.0](https://github.com/1Password/terraform-provider-onepassword/releases/tag/v3.0.0) we migrated to use the SDK instead!

The error you are getting here seems more related to the item UUID:
```
│ Unable to read item, got error: failed to get item using sdk: invalid user
│ input: encountered the following errors: the provided ID is not a valid
│ UUID
```
It looks like you were possibly able to authenticate with Service Account but the item ID was an invalid UUID. 🤔 

Are you able to share more of the provider config you used so we can help dig into why you may be receiving this error?

dbrennand:
Hi @JillRegan 

Thank you for your prompt response 🙂 

```
data "onepassword_item" "ssh_key" {
  vault = var.op_ssh_vault_name
  title = var.op_ssh_key_name
}
```

[...]

PR #307, opened 2026-01-05
Get vault UUID for item data source

JillRegan:
### ✨ Summary

- When providing the vault name to try and get an item the provider was throwing an error as the SDK expects a UUID instead of a vault name. 
- This was a regression from migrating to SDKs.
- To fix this when we get an item for data source we will check if the vault UUID is valid first. If not we will make a get request for the vault using the name in order to get the UUID

### 🔗 Resolves:

Resolves: https://github.com/1Password/terraform-provider-onepassword/issues/306

### ✅ Checklist

- [x] 🖊️ Commits are signed
  - [ ] 🔹 Unit /🔸 Integration
  - [x] 🌐 E2E

You can reproduce the bug but trying to get an item using the vault name using SDK:

```
data "onepassword_item" "test" {
  vault    = "terraform-provider-acceptance-tests"
  title    = "Test API Credential"
}
```

### DIFF
--- a/internal/onepassword/sdk/client.go
+++ b/internal/onepassword/sdk/client.go
@@ -246,21 +246,21 @@ func NewClient(ctx context.Context, config SDKConfig) (*Client, error) {
 }
 
 // resolveVaultUUID resolves a vault name to a UUID
-func (c *Client) resolveVaultUUID(ctx context.Context, vaultValue string) (string, error) {
-	if util.IsValidUUID(vaultValue) {
-		return vaultValue, nil
+func (c *Client) resolveVaultUUID(ctx context.Context, vaultQuery string) (string, error) {
+	if util.IsValidUUID(vaultQuery) {
+		return vaultQuery, nil
 	}
 
 	// Vault value is a name, resolve it to UUID
-	vaults, err := c.GetVaultsByTitle(ctx, vaultValue)
+	vaults, err := c.GetVaultsByTitle(ctx, vaultQuery)
 	if err != nil {
 		return "", fmt.Errorf("failed to get vault by title: %w", err)
 	}
 	if len(vaults) == 0 {
-		return "", fmt.Errorf("no vault found with name %q", vaultValue)
+		return "", fmt.Errorf("no vault found with name %q", vaultQuery)
 	}
 	if len(vaults) > 1 {
-		return "", fmt.Errorf("multiple vaults found with name %q", vaultValue)
+		return "", fmt.Errorf("multiple vaults found with name %q", vaultQuery)
 	}
 	return vaults[0].ID, nil
 }


test/e2e/item_data_source_test.go
```diff
@@ -94,7 +94,7 @@ var testItems = map[model.ItemCategory]testItem{
 	},
 	model.APICredential: {
 		Title: "Test API Credential",
-		UUID:  "pkqk4ehxuqp4fy3wpjep5qozru",
+		UUID:  "fmx3tsd5v5bew74o3v2z2a7wf4",
```
JillRegan, line 97: Needed to update as I was updating test data in another PR and end up removing the wrong API Credential.

internal/provider/onepassword_item_data_source.go
```diff
@@ -425,10 +426,30 @@ func (d *OnePasswordItemDataSource) Read(ctx context.Context, req datasource.Rea
...
+	vaultValue := data.Vault.ValueString()
 	itemTitle := data.Title.ValueString()
 	itemUUID := data.UUID.ValueString()
 
+	// Resolve vault name to UUID if needed
+	var vaultUUID string
+	if util.IsValidUUID(vaultValue) {
+		// Vault value is a UUID, use it directly
+		vaultUUID = vaultValue
+	} else {
```
volodymyrZotov: I see that [connect-sdk-go](https://github.com/1Password/connect-sdk-go/blob/main/connect/client.go#L231) handles vault ID check internally.

I'd move vault ID check into `sdkClient.GetItemByTitle` and `sdkClient.GetItem` and leave this function unchanged. This way we make sure both clients (SDK and Connect) behave the same way.

WDYT?

JillRegan: Good point, since Connect handles this internally we might as well make this change specific to SDK. I have updated. 👍

internal/onepassword/sdk/client.go
```diff
@@ -232,3 +244,23 @@ func NewClient(ctx context.Context, config SDKConfig) (*Client, error) {
 		config:    config,
 	}, nil
 }
+
+// resolveVaultUUID resolves a vault name to a UUID
+func (c *Client) resolveVaultUUID(ctx context.Context, vaultValue string) (string, error) {
```
volodymyrZotov: `[nit]`

`vaultValue` could be renamed to `vaultQuery` to align with the connect-sdk-go implementation.

test/e2e/item_data_source_test.go
```diff
@@ -445,3 +445,62 @@ func TestAccItemDataSource_DetectManualChanges(t *testing.T) {
...
+		name     string
+	}{
+		{model.Login, "Login"},
+	}
+
+	var testCases []itemDataSourceTestCase
+
+	// Create test cases for each item type with both title and UUID lookup methods
+	for _, itemType := range itemTypes {
+		item := testItems[itemType.category]
```
volodymyrZotov: Seems this is wrong here, as `itemTypes` is a slice.

PR #309, opened 2026-01-07
Fields and sections map for item resource

volodymyrZotov:
### ✨ Summary

This PR introduces `section_map` as an alternative to the existing `section (list)` attribute for managing custom sections and fields in the `onepassword_item_resource`.

#### Key changes:
- Added `section_map` attribute that uses section labels as map keys for direct lookup.
- Each section contains `field_map` where field labels are used as keys.
- Fields support `type`, `value`, `id`, and `password_recipe` for password generation.
- Added mutual exclusivity validation - users cannot use both `section (list)` and `section_map` at the same time within same resource.
- Terraform's native map semantics enforce unique keys (duplicate keys result in last value winning).
- Created separate description constants for resource and data source to avoid misleading docs.

#### Benefits:
- Direct access to sections/fields by label: `section_map["my_section"].field_map["api_key"].value`.
- Cleaner syntax for managing structured credentials.
- Type-safe field definitions with optional password generation.

### ✅ Checklist

- [x] 🖊️ Commits are signed
  - [x] 🔹 Unit /🔸 Integration
  - [x] 🌐 E2E

`section_map` is only available on the resource. Data source to be addressed separately.

### DIFF
--- a/internal/provider/state_to_model.go
+++ b/internal/provider/state_to_model.go
@@ -243,15 +243,13 @@ func toModelSectionFieldMap(field OnePasswordItemResourceFieldMapModel, fieldLab
 		fieldID = sid
 	}
 
-	fieldValue := field.Value.ValueString()
 	modelItemField := &model.ItemField{
 		SectionID:    sectionID,
 		SectionLabel: sectionLabel,
 		ID:           fieldID,
 		Type:         model.ItemFieldType(op.ItemFieldType(field.Type.ValueString())),
-		Label:        fieldLabel, // Use the map key as the label
-		Value:        fieldValue,
-		Generate:     fieldValue == "", // Generate password if value is empty
+		Label:        fieldLabel,
+		Value:        field.Value.ValueString(),
 	}
 
 	recipe, err := parseGeneratorRecipeFromModel(field.Recipe)
--- a/internal/provider/state_to_model_test.go
+++ b/internal/provider/state_to_model_test.go
@@ -675,3 +675,90 @@ func TestToModelTags(t *testing.T) {
 		})
 	}
 }
+
+func TestToModelSectionFieldMap(t *testing.T) {
+	tests := map[string]struct {
+		field        OnePasswordItemResourceFieldMapModel
+		fieldLabel   string
+		sectionID    string
+		sectionLabel string
+		wantErr      bool
+		validate     func(t *testing.T, field *model.ItemField)
+	}{
+		"with existing field ID": {
+			field: OnePasswordItemResourceFieldMapModel{
+				ID:    types.StringValue("existing-field-id"),
+				Type:  types.StringValue("STRING"),
+				Value: types.StringValue("test value"),
+				Recipe: &PasswordRecipeModel{
+					Length:  types.Int64Value(16),
+					Digits:  types.BoolValue(true),
+					Symbols: types.BoolValue(true),
+				},
+			},
+			fieldLabel:   "Test Field",
+			sectionID:    "section-id",
+			sectionLabel: "Section Label",
+			wantErr:      false,
+			validate: func(t *testing.T, field *model.ItemField) {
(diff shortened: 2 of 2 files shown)


internal/provider/state_to_model.go
```diff
@@ -199,3 +199,103 @@ func toModelTags(ctx context.Context, state OnePasswordItemResourceModel) ([]str
 	}
 	return tags, nil
 }
+
+func parseGeneratorRecipeFromModel(recipe *PasswordRecipeModel) (*model.GeneratorRecipe, error) {
```
JillRegan, line 203: Are we able to replace `parseGeneratorRecipe` with this method? They look almost identical.

volodymyrZotov: Updated

internal/provider/state_to_model.go
```diff
@@ -199,3 +199,103 @@ func toModelTags(ctx context.Context, state OnePasswordItemResourceModel) ([]str
...
+
+	fieldValue := field.Value.ValueString()
+	modelItemField := &model.ItemField{
+		SectionID:    sectionID,
+		SectionLabel: sectionLabel,
+		ID:           fieldID,
+		Type:         model.ItemFieldType(op.ItemFieldType(field.Type.ValueString())),
+		Label:        fieldLabel, // Use the map key as the label
+		Value:        fieldValue,
+		Generate:     fieldValue == "", // Generate password if value is empty
```
JillRegan: Should we check if a password recipe exists first? We allow empty field values to be set and I think this may trigger Connect to generate a password for it. For SDK though I believe we handle it with so that would be okay.
```
	if f.Generate && f.Recipe != nil {
		password, err := generatePassword(f.Recipe)
```

JillRegan: Just tested with this config using Connect:

```
resource "onepassword_item" "test_generate_with_recipe" {
  vault    = "bbucuyq2nn4fozygwttxwizpcy"
  title    = "Test Generate With Recipe test"
  category = "login"

  section_map = {
    "test_section" = {
      field_map = {
        "should_generate" = {
          type  = "CONCEALED"
          value = ""

        }
      }
    }
  }
}
```

And a password was generated

```
 "section_map": {
              "test_section": {
                "field_map": {
                  "should_generate": {
                    "id": "a9852e3a-1059-2fd9-adf9-1a86131ef16e",
                    "password_recipe": null,
                    "type": "CONCEALED",
                    "value": "!CNptYR6rBnB38qnK--UNeMx7kG*efCR"
                  }
                },
                "id": "b7acb6aa-6966-f4ed-6244-5a1b4a43f4b1"
 }
```

docs/data-sources/item.md
```diff
@@ -38,15 +38,15 @@ data "onepassword_item" "example" {
 - `category` (String) The category of the item. One of ["login" "password" "database" "secure_note" "document" "ssh_key"]
 - `credential` (String, Sensitive) API credential for this item.
 - `database` (String) (Only applies to the database category) The name of the database.
-- `file` (Block List) A list of files attached to the item. (see [below for nested schema](#nestedblock--file))
+- `file` (Block List) A list of files attached to the document item. (see [below for nested schema](#nestedblock--file))
```
JillRegan, line 41: I think the file doesn't just have to be on the document item. It can also be added to other items.

internal/provider/state_to_model.go
volodymyrZotov: Removed `Generate` property. So when empty value is provided we do not generate password. That make sense as field can contain a simple `STRING` value not necessary a password value.

docs/data-sources/item.md
volodymyrZotov: There is some confusion there. In datasource the top level `file` is a file from Document item.

All other that are attached go under `section.file[x]`.

I decided to update description to make it a bit clearer what file you should expect at the top level.

Issue #311, opened 2026-01-07
`MONTH_YEAR` field type has different format for Connect and SDK


volodymyrZotov:
### Your environment

Terraform Provider Version: 3.0.2

Authentication Method: all

Connect Server Version: 1.7.2

OS: Macos

Terraform Version: 1.14.0

## What did you expect to happen?

`MONTH_YEAR` field type has different format for Connect and SDK

- SDK -> `MM/YYYY`
- Connect -> `YYYYMM` or `YYYY/MM`

That leads to the issue when switching the auth method.

## Steps to reproduce

1. Create a field with type `MONTH_YEAR` and value `01/2025` using Connect API
2. Observe error: `"values for field type monthYear must be in YYYYMM or YYYY/MM format"`
3. Create the same field using Service Account SDK - it succeeds
4. Reverse: use `202501` format - Connect succeeds, SDK may have issues

PR #313, opened 2026-01-07
Ensure YYYYMM is passed for MONTH_YEAR field

JillRegan:
### ✨ Summary

- This PR adds a validator to ensure we pass YYYYMM value for the `MONTH_YEAR` field. 
- Connect accepts YYYYMM or YYYY/MM, but it always returns back YYYYMM. To keep consistent with this I have added a validator to ensure we continue to only pass YYYYMM.
- For SDK it needs to be converted to MM/YYYY

### 🔗 Resolves:

Resolves: https://github.com/1Password/terraform-provider-onepassword/issues/311

### ✅ Checklist

- [x] 🖊️ Commits are signed
  - [x] 🔹 Unit /🔸 Integration
  - [x] 🌐 E2E

### DIFF
--- a/internal/provider/month_year_validator.go
+++ b/internal/provider/month_year_validator.go
@@ -3,13 +3,11 @@ package provider
 import (
 	"context"
 	"fmt"
-	"regexp"
+	"time"
 
 	"github.com/hashicorp/terraform-plugin-framework/schema/validator"
 	"github.com/hashicorp/terraform-plugin-framework/types"
 )
-
-var monthYearRegex = regexp.MustCompile(`^\d{4}(0[1-9]|1[0-2])$`)
 
 func validateMonthYear() monthYearValidator {
 	return monthYearValidator{}
@@ -42,7 +40,8 @@ func (v monthYearValidator) ValidateString(ctx context.Context, req validator.St
 
 	value := req.ConfigValue.ValueString()
 
-	if !monthYearRegex.MatchString(value) {
+	_, err := time.Parse("200601", value)
+	if err != nil {
 		resp.Diagnostics.AddAttributeError(
 			req.Path,
 			"Invalid MONTH_YEAR format",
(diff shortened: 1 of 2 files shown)


internal/provider/month_year_validator.go
```diff
@@ -0,0 +1,52 @@
...
+
+type monthYearValidator struct{}
+
+func (v monthYearValidator) Description(ctx context.Context) string {
+	return "MONTH_YEAR values must be in YYYYMM format (e.g., 202401)"
+}
+
+func (v monthYearValidator) MarkdownDescription(ctx context.Context) string {
+	return "MONTH_YEAR values must be in YYYYMM format (e.g., 202401)"
+}
```
JillRegan, line 24: While Connect accepts both YYYYMM and YYYY/MM formats, it will always return YYYYMM. Current behaviour will allow user to use YYYY/MM but provider will throw and error after apply as there are state inconsistencies as the values don't match. 

To avoid these state inconsistencies I think it is best if we enforce users only use YYYYMM date format for this field.

internal/provider/onepassword_item_resource.go
```diff
@@ -397,6 +397,7 @@ func (r *OnePasswordItemResource) Schema(ctx context.Context, req resource.Schem
 											stringvalidator.ConflictsWith(
 												path.MatchRelative().AtParent().AtName("password_recipe"),
 											),
+											validateMonthYear(),
```
volodymyrZotov, line 401: Let's add this validator for `field_map.value` as well.

internal/provider/month_year_validator.go
volodymyrZotov: makes sense

internal/provider/month_year_validator.go
```diff
@@ -0,0 +1,52 @@
...
+		return
+	}
+
+	if fieldType.ValueString() != "MONTH_YEAR" {
+		return
+	}
+
+	value := req.ConfigValue.ValueString()
+
+	if !monthYearRegex.MatchString(value) {
```
volodymyrZotov: `[nit]`

Though, regexp works fine here, I'd rather use `time` standard library package to parse. I think it's easier to read

```
_, err := time.Parse("200601", value)
```

PR #317, opened 2026-01-12
Ensure test vault Id syncs before running tests

JillRegan:
### ✨ Summary

- Now that all tests run in parallel there seemed to be an issue an issue with the test Vault not being provided. This was due to GetTestVaultID querying multiple times simultaneously and then caching an empty response. See [here](https://github.com/1Password/terraform-provider-onepassword/actions/runs/20935893012/job/60158298967?pr=313). 
- To fix this I have added a TestMain to initialize the vault before running the tests. 

### ✅ Checklist

- [x] 🖊️ Commits are signed
  - [ ] 🔹 Unit /🔸 Integration
  - [x] 🌐 E2E

### DIFF
--- a/test/e2e/item_resource_test.go
+++ b/test/e2e/item_resource_test.go
@@ -109,10 +109,13 @@ var testItemsUpdatedAttrs = map[model.ItemCategory]map[string]any{
 }
 
 func TestMain(m *testing.M) {
-	// Initialize early if using custom vault (prevents race conditions in parallel tests)
+	// Initialize vault early if using custom vault (prevents race conditions in parallel tests)
 	// for default vault, this is instant as we already have the ID
-	vault.GetTestVaultID(nil)
-
+	err := vault.InitTestVaultID()
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "failed to initialize test vault: %v\n", err)
+		os.Exit(1)
+	}
 	code := m.Run()
 	os.Exit(code)
 }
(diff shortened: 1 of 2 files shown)


test/e2e/utils/vault/vault.go
```diff
@@ -14,12 +14,15 @@ var (
 	testVaultID     string
 )
 
-// GetTestVaultID returns the vault ID by querying by name once and caching the result.
+const defaultTestVaultID = "bbucuyq2nn4fozygwttxwizpcy"
```
JillRegan, line 19: Majority of the time we are using the provider test vault and not setting OP_TEST_VAULT_NAME for local runs. To save a little time I think if OP_TEST_VAULT_NAME is not set we just return the default ID instead of adding another query.

test/e2e/item_resource_test.go
```diff
@@ -107,6 +108,15 @@ var testItemsUpdatedAttrs = map[model.ItemCategory]map[string]any{
 	},
 }
 
+func TestMain(m *testing.M) {
+	// Initialize early if using custom vault (prevents race conditions in parallel tests)
+	// for default vault, this is instant as we already have the ID
+	vault.GetTestVaultID(nil)
```
volodymyrZotov: When passing `nil` the function will panic in case `t` will be accessed.

test/e2e/utils/vault/vault.go
volodymyrZotov: If passing `t` as `nil` this line and others that use `t` will panic.

JillRegan: Ahh yes. I have updated to create a separate vault initialization method in order to ensure running t.Fatalf for errors only happens in actual test cases.

volodymyrZotov (approved):

Issue #310, opened 2026-01-07
Failure to find vault by name or UUID


mike-prime:
### Your environment

Terraform Provider Version: 3.0.2

Authentication Method: macOS app

Connect Server Version: N/A

OS: 26.2

Terraform Version: v1.12.1

## What happened?

Could not find a vault that exists and my account has access to.

## What did you expect to happen?

Found the vault by name or UUID.

## Steps to reproduce

1. Try and execute a plan with a data source like this:

```hcl
data "onepassword_vault" "this" {
  name = "Infrastructure"
}
```

2. Observe the error `No vault found with name '"Infrastructure"'`

3. Attempt to use UUID:

```hcl
data "onepassword_vault" "this" {
  # name = "Infrastructure"
  uuid = "correctuuidhere"
}
```

4. Observe the error `failed to get vault using sdk: data did not match any variant of untagged enum Invocation at line 1 column 129`

## Notes & Logs

If I use the vault's UUID directly in an item data source it works fine. I've tried this in all released 3.0.x versions and its the same behavior. One thing that stands out to me as somewhat odd is the extra quotes around the vault name: `No vault found with name '"Infrastructure"'`.

Debug error from name query:

```
2026-01-07T14:19:05.201-0500 [ERROR] provider.terraform-provider-onepassword_v3.0.2: Response contains error diagnostic: diagnostic_detail="No vault found with name '\"Infrastructure\"'" diagnostic_severity=ERROR tf_data_source_type=onepassword_vault tf_proto_version=6.10 tf_req_id=9020d490-67c1-66ea-6576-08765a9f0e38 tf_rpc=ReadDataSource @caller=github.com/hashicorp/terraform-plugin-go@v0.29.0/tfprotov6/internal/diag/diagnostics.go:58 @module=sdk.proto diagnostic_summary="Client Error" tf_provider_addr=registry.terraform.io/1password/onepassword timestamp=2026-01-07T14:19:05.201-0500
```

Debug error from UUID query:

```
[...]

PR #323, opened 2026-01-19
Add temporary fallback to GetVault 

JillRegan:
### ✨ Summary

When trying to access vault directly with desktop auth error is returned:
```
failed to get vault using sdk: data did not match any
│ variant of untagged enum Invocation at line 1 column 129
```
Recent changes to beta SDK result in this new error. A fix will be shipped before stable release but until then we should add this fallback to list all vaults and extract one we want to unblock Provider users.

### 🔗 Resolves:

Resolves: https://github.com/1Password/terraform-provider-onepassword/issues/310

### ✅ Checklist

- [x] 🖊️ Commits are signed
  - [ ] 🔹 Unit /🔸 Integration
  - [ ] 🌐 E2E

To reproduce this bug attempt to access a vault directly using the SDK with OP_ACCOUNT with name or uuid. It will produce an error. 

```
data "onepassword_vault" "test_vault" {
  #name = <your-vault-name>
  uuid = <your-vault-uuid>
}
```

With the changes here we will fallback to working `Vault().List()` in order to get vault info.

### DIFF
--- a/internal/onepassword/sdk/client.go
+++ b/internal/onepassword/sdk/client.go
@@ -32,7 +32,7 @@ func (c *Client) GetVault(ctx context.Context, uuid string) (*model.Vault, error
 			decryptDetails := true
 			vaultList, listErr := c.sdkClient.Vaults().List(ctx, sdk.VaultListParams{DecryptDetails: &decryptDetails})
 			if listErr != nil {
-				return nil, fmt.Errorf("failed to get vault using sdk: %w (fallback list also failed: %v)", err, listErr)
+				return nil, fmt.Errorf("failed to get vault using sdk: %w (fallback list failed: %v)", err, listErr)
 			}
 
 			// Find vault with matching UUID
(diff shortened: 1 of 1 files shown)


internal/onepassword/sdk/client.go
```diff
@@ -25,7 +25,29 @@ type SDKConfig struct {
...
+					result := &model.Vault{}
+					result.FromSDKVault(&v)
+					return result, nil
+				}
+			}
+
+			return nil, fmt.Errorf("failed to get vault using sdk: %w (vault with UUID %q not found in fallback list)", err, uuid)
+		}
+
+		return nil, fmt.Errorf("failed to get vault using sdk: %w (vault with UUID %q not found in fallback list)", err, uuid)
```
volodymyrZotov: Maybe leave the same message as it was previously?

```suggestion
		return nil, fmt.Errorf("failed to get vault using sdk: %w", err)
```

internal/onepassword/sdk/client.go
```diff
@@ -25,7 +25,29 @@ type SDKConfig struct {
...
 	if err != nil {
-		return nil, fmt.Errorf("failed to get vault using sdk: %w", err)
+		// TODO: Remove this fallback once the SDK is stable
+		// This is a temporary workaround to a bug in the SDK when accessing vault directly by uuid
+		errStr := err.Error()
+		if strings.Contains(errStr, "enum Invocation") {
+			decryptDetails := true
+			vaultList, listErr := c.sdkClient.Vaults().List(ctx, sdk.VaultListParams{DecryptDetails: &decryptDetails})
+			if listErr != nil {
+				return nil, fmt.Errorf("failed to get vault using sdk: %w (fallback list also failed: %v)", err, listErr)
```
volodymyrZotov:
```suggestion
				return nil, fmt.Errorf("failed to get vault using sdk: %w (fallback list failed: %v)", err, listErr)
```

PR #328, opened 2026-01-22
Add ephemeral item resource provider

BlackHoleFox:
### ✨ Summary

Attempts an initial implementation at adding support for [ephemeral resource blocks](https://developer.hashicorp.com/terraform/language/manage-sensitive-data/ephemeral) for the `onepassword_item` resource. The new resource is heavily based off `onepassword_item_data_source.go` since its really just a read-only set of values in a different context.

Apologies in advance since my Golang knowledge is very absent though (learned what an interface was tonight).

I intentionally didn't mirror the full API surface of the item data source or resource for the ephemeral version. Given the limited contexts ephemeral blocks can be used in (primarily for configuring providers outside of WO fields) it didn't seem item values that aren't useful for logging in would be that relevant.

Tested successfully with both `opentofu` and `terraform` on macOS. Values are fetched identically to the data source from the user's perspective but nothing ends up in the state file.

### 🔗 Resolves:

#212

### ✅ Checklist

- [x] 🖊️ Commits are signed
  - [x] 🔹 Unit /🔸 Integration
  - [x] 🌐 E2E

Due to unfamiliarity with Go I have not the foggiest where to start writing tests or what's worth testing. Logic is entirely reused from the existing resource types but has no tests of its own for the time. Reviewer assistance appreciated if possible :D

### DIFF
--- a/internal/provider/onepassword_item_data_source.go
+++ b/internal/provider/onepassword_item_data_source.go
@@ -141,7 +141,7 @@ func (d *OnePasswordItemDataSource) Schema(ctx context.Context, req datasource.S
 				Required:            true,
 			},
 			"uuid": schema.StringAttribute{
-				MarkdownDescription: itemProvidingTypeUUIDDescription,
+				MarkdownDescription: itemLookupUUIDDescription,
 				Optional:            true,
 				Computed:            true,
 				Validators: []validator.String{
@@ -152,7 +152,7 @@ func (d *OnePasswordItemDataSource) Schema(ctx context.Context, req datasource.S
 				},
 			},
 			"title": schema.StringAttribute{
-				MarkdownDescription: itemProvidingTypeTitleDescription,
+				MarkdownDescription: itemLookupTitleDescription,
 				Optional:            true,
 				Computed:            true,
 			},
--- a/internal/provider/onepassword_item_ephemeral.go
+++ b/internal/provider/onepassword_item_ephemeral.go
@@ -68,7 +68,7 @@ func (r *OnePasswordItemEphemeral) Schema(ctx context.Context, req ephemeral.Sch
 				Required:            true,
 			},
 			"uuid": schema.StringAttribute{
-				MarkdownDescription: itemProvidingTypeUUIDDescription,
+				MarkdownDescription: itemLookupUUIDDescription,
 				Optional:            true,
 				Computed:            true,
 				Validators: []validator.String{
@@ -79,7 +79,7 @@ func (r *OnePasswordItemEphemeral) Schema(ctx context.Context, req ephemeral.Sch
 				},
 			},
 			"title": schema.StringAttribute{
-				MarkdownDescription: itemProvidingTypeTitleDescription,
+				MarkdownDescription: itemLookupTitleDescription,
 				Optional:            true,
 				Computed:            true,
 			},
--- a/internal/provider/const.go
+++ b/internal/provider/const.go
@@ -12,8 +12,8 @@ const (
 	itemDataSourceDescription = "Use this to get details of an item by its vault uuid and either the title or the uuid of the item."
 	itemEphemeralDescription  = "Use this to retrieve item values without storing them in Terraform state. Useful for providing sensitive values to write-only arguments or other ephemeral contexts."
 
-	itemProvidingTypeUUIDDescription  = "The UUID of the item to retrieve. This field will be populated with the UUID of the item if the item it looked up by its title."
-	itemProvidingTypeTitleDescription = "The title of the item to retrieve. This field will be populated with the title of the item if the item it looked up by its UUID."
+	itemLookupUUIDDescription  = "The UUID of the item to retrieve. This field will be populated with the UUID of the item if the item it looked up by its title."
+	itemLookupTitleDescription = "The title of the item to retrieve. This field will be populated with the title of the item if the item it looked up by its UUID."
 
 	itemUUIDDescription                  = "The UUID of the item. Item identifiers are unique within a specific vault."
 	vaultUUIDDescription                 = "The UUID of the vault the item is in."


docs/resources/item.md
```diff
@@ -81,15 +81,17 @@ resource "onepassword_item" "example_with_list" {
 
 ### Optional
 
+> **NOTE**: [Write-only arguments](https://developer.hashicorp.com/terraform/language/resources/ephemeral#write-only-arguments) are supported in Terraform 1.11 and later.
+
 - `category` (String) The category of the item. One of ["login" "password" "database" "secure_note"]
```
JillRegan, line 86: I had to update terraform-plugin-docs to v0.24.0 in order to include new ephemeral resource item in the docs. This also made some additions to resource items documentation for the WO values.

JillRegan:
@BlackHoleFox Thanks so much for this, this looks great! 🔥 

I have added both Unit tests and E2E tests for this PR for the ephemeral item. I also made adjustments to the docs in order to include this new item. 😄 

I am just going to tag another member of our team in for review, as now that I have contributed it is best practice to have another review.

internal/provider/onepassword_item_data_source.go
```diff
@@ -520,11 +520,7 @@ func (d *OnePasswordItemDataSource) Read(ctx context.Context, req datasource.Rea
 	}
 }
 
-func getItemForDataSource(ctx context.Context, client onepassword.Client, data OnePasswordItemDataSourceModel) (*model.Item, error) {
-	vaultUUID := data.Vault.ValueString()
-	itemTitle := data.Title.ValueString()
-	itemUUID := data.UUID.ValueString()
-
+func getItemForProvidingType(ctx context.Context, client onepassword.Client, vaultUUID string, itemTitle string, itemUUID string) (*model.Item, error) {
```
volodymyrZotov: Function name sound confusing. Not sure that I understand what's the purpose of this function reading this new function name.

Would it make sense to call it `getItem`? It seems clearer explains what the function suppose to do?

internal/provider/onepassword_item_resource_ephemeral.go
```diff
@@ -0,0 +1,206 @@
...
+	return &OnePasswordEphemeralItemResource{}
+}
+
+// OnePasswordEphemeralItemResource defines the ephemeral resource implementation.
+type OnePasswordEphemeralItemResource struct {
+	client onepassword.Client
+}
+
+// OnePasswordEphemeralItemResourceModel describes the data source data model.
+type OnePasswordEphemeralItemResourceModel struct {
```
volodymyrZotov: What's the reason to use both `Ephemeral` and `Resource` in the name? `resource` means it controls full lifecycle of an item (supports CRUD). But in this case it only reads and item.

Would it make sense to rename?
```suggestion
type OnePasswordEphemeralItemModel struct {
```

internal/provider/onepassword_item_ephemeral.go
```diff
@@ -0,0 +1,206 @@
...
+	Title             types.String `tfsdk:"title"`
+	URL               types.String `tfsdk:"url"`
+	Hostname          types.String `tfsdk:"hostname"`
+	Username          types.String `tfsdk:"username"`
+	Password          types.String `tfsdk:"password"`
+	NoteValue         types.String `tfsdk:"note_value"`
+	Credential        types.String `tfsdk:"credential"`
+	PublicKey         types.String `tfsdk:"public_key"`
+	PrivateKey        types.String `tfsdk:"private_key"`
+	PrivateKeyOpenSSH types.String `tfsdk:"private_key_openssh"`
```
volodymyrZotov, line 49: The model doesn't support files. This is fine. And files support can be added in the separate PR. Would be good to create an issue to track this.

internal/provider/onepassword_item_ephemeral.go
```diff
@@ -0,0 +1,206 @@
...
+	client onepassword.Client
+}
+
+// OnePasswordEphemeralItemResourceModel describes the data source data model.
+type OnePasswordEphemeralItemResourceModel struct {
+	Vault             types.String `tfsdk:"vault"`
+	UUID              types.String `tfsdk:"uuid"`
+	Title             types.String `tfsdk:"title"`
+	URL               types.String `tfsdk:"url"`
+	Hostname          types.String `tfsdk:"hostname"`
```
volodymyrZotov, line 39: As we adding a field from `Database` item category in this PR, I'd also add other properties as well `Port`, `Database`,  `Type`, so we support the same properties as item datasource.

That will match the users expectations (aka datasource supports those Database item fields, so ephemeral might support the same).

Another option, is to remove `Hostname` from this PR and file a separate issue `Add Database item category support for ephemeral items` where we can implement all fields together.

That will also help users to understand what's not supported, what exactly should be added. Also allowing contributors to implement it.

internal/provider/onepassword_item_ephemeral.go
```diff
@@ -0,0 +1,206 @@
...
+	Username          types.String `tfsdk:"username"`
+	Password          types.String `tfsdk:"password"`
+	NoteValue         types.String `tfsdk:"note_value"`
+	Credential        types.String `tfsdk:"credential"`
```
volodymyrZotov, line 46: Similar to the comment about `Database` category.
I'd either add `ValidFrom` to make it support the same fields for `Credentials` item category as item datasource, or create an issue and for now remove.

internal/provider/onepassword_item_resource_ephemeral.go
volodymyrZotov: Ideally I'd expect this model has the same properties as datasource. Maybe except of SectionMap and SectionList. 

I see that this PR intentionally doesn't implement all of them, and this is fine.

Let's create the issues (maybe several) so missing fields can be added in separate PRs.

This will give us clear picture what fields are missing and give opportunity for contributors to help with that.

internal/provider/onepassword_item_ephemeral.go
```diff
@@ -0,0 +1,206 @@
...
+func (r *OnePasswordEphemeralItemResource) Metadata(ctx context.Context, req ephemeral.MetadataRequest, resp *ephemeral.MetadataResponse) {
+	resp.TypeName = req.ProviderTypeName + "_item"
+}
+
+func (r *OnePasswordEphemeralItemResource) Schema(ctx context.Context, req ephemeral.SchemaRequest, resp *ephemeral.SchemaResponse) {
+	resp.Schema = schema.Schema{
+		// This description is used by the documentation generator and the language server.
+		MarkdownDescription: itemEphemeralResourceDescription,
+
+		Attributes: map[string]schema.Attribute{
```
volodymyrZotov, line 61: `id` attribute is missing, let's add it. Use the same logic to populate as we use in datasource.

Also `ID` property should be added to the model.

internal/provider/onepassword_item_resource_ephemeral.go
```diff
@@ -0,0 +1,206 @@
...
+					}
+					data.PrivateKeyOpenSSH = types.StringValue(openSSHPrivateKey)
+				case "credential":
+					data.Credential = types.StringValue(f.Value)
+				}
+			}
+		}
+	}
+
+	tflog.Trace(ctx, "read an item data source")
```
volodymyrZotov: Any benefits of having this log? Maybe remove it?
Or at least change the message

```suggestion
	tflog.Trace(ctx, "read an ephemeral item")
```

internal/provider/onepassword_item_ephemeral.go
BlackHoleFox: `Hostname` also comes from the `Credential` category type, so maybe a followup for the whole Database type?

internal/provider/onepassword_item_resource_ephemeral.go
BlackHoleFox: > What's the reason to use both Ephemeral and Resource in the name? 

Can change this if reviewers would like but the Terraform docs refer to them as "ephemeral resources" so that seemed easiest to follow. Other official Terraform plugins also use the same naming scheme.

internal/provider/onepassword_item_ephemeral.go
BlackHoleFox: What's the usecase for `ValidFrom` in an ephemeral context?

internal/provider/onepassword_item_resource_ephemeral.go
volodymyrZotov: I think the the terraform docs them mention `resource` as they need to describe what is `ephemeral` and that's appropriate for documentation and user-facing content.

However, for internal struct names, having `ItemResource`, `ItemDataSource`, and `ItemEphemeral` makes a very clear distinction between all of them and follows the existing pattern in this codebase. (aka we already use `ItemDataSource` and not `ItemDataSourceResource`).

In a contrast `ItemEphemeralResource` and `ItemResource`, may bring a bit of confusion, especially for those new contributors who are not very familiar with terraform resource types and their lifecycles.

internal/provider/onepassword_item_ephemeral.go
volodymyrZotov: As provider already supports `ValidFrom` on datasource. I imagine some users might want to convert datasource into ephemeral, and therefore they might expect to see `ValidFrom` on ephemeral as well.

internal/provider/onepassword_item_ephemeral.go
volodymyrZotov: I'm ok with either direction.

My whole point, is to not partially implement support for the categories in this PR, so it's easier to create separate issues for a specific category.

Aka having only `Hostname`, which will cover both `Credentials` and `Database` item categories, but neither of them have all list of available values like `ValidFrom`, `Type`, `Port` etc.

internal/provider/onepassword_item_resource_ephemeral.go
JillRegan: I think even though this is technically called an `Ephemeral Resource `and think it makes sense internally to just refer to it as `Ephemeral`. In order to avoid any confusion in the future. I have updated 👍

internal/provider/onepassword_item_ephemeral.go
JillRegan: Makes sense, issue created here: https://github.com/1Password/terraform-provider-onepassword/issues/330

internal/provider/onepassword_item_ephemeral.go
JillRegan: I have added handling for type, port, and datebase fields but for now have skipped validFrom.

internal/provider/onepassword_item_ephemeral.go
JillRegan: Added 👍

internal/provider/onepassword_item_resource_ephemeral.go
JillRegan: Removed 👍

Issue #312, opened 2026-01-07
Password is updated after on imported resource


volodymyrZotov:
### Your environment

Terraform Provider Version: 3.0.2

Authentication Method: service accounts

Connect Server Version: n/a

OS: macos

Terraform Version: 1.14.0

## What did you expect to happen?
After importing resource and running `terraform apply` the password is updated.

## Steps to reproduce

1. Create a Terraform configuration with password_recipe:
```
resource "onepassword_item" "example" {
  vault = "your-vault-id"
  title = "Test Item"
  category = "login"

  password_recipe {
    length  = 32
    digits  = true
    symbols = true
  }
}
```
2. Run `terraform apply` to create the item
3. Remove the resource from state: `terraform state rm onepassword_item.example`
4. Import the resource: `terraform import onepassword_item.example vaults/<vault-id>/items/<item-id>`
5. Run terraform apply
6. Observe that the password value has changed

PR #335, opened 2026-02-10
Preserve password with recipe after import

JillRegan:
### ✨ Summary

After importing an item that was created with `password_recipe` the next terraform apply regenerated the password even though the config was unchanged. On import state is filled from the 1Password. This does not return the generator recipe, so after import state had no `password_recipe`. The value modifier only preserved the password when state and plan recipes were equal, so it treated import and then terraform apply as a change and did not preserve the password.

The value modifier used for top-level and section list password fields now treats the post-import case as recipe unchanged and preserves the existing password:

- Recipes equal: preserve (unchanged behaviour).
- State has no recipe, plan has recipe: preserve (new: post-import).
- Recipes differ: do not preserve (unchanged behaviour).

### Note on `section_map` as this is not addressed here

After import, state is only populated with the section list from the API, not section_map. 

```
	if len(state.SectionMap) > 0 {
		state.SectionMap = toStateSectionsAndFieldsMap(modelItem, state.SectionMap)
	} else {
		state.SectionList = toStateSectionsAndFieldsList(modelItem.Sections, modelItem.Fields, state.SectionList)
	}
```

After import, section_map is not populated because at Read we don’t have the information needed to fill it: we only have prior state (minimal after import) and no config. So we can’t know the user used section_map, and in modelToState we only populate one of the two; with both empty we default to populating the section list only.

We do not populate both section and section_map in state because a resource item can only have one in config (either section or section_map). Populating both would make state inconsistent with that and with how the resource is defined, so we keep a single representation in state.

### 🔗 Resolves:

Resolves: https://github.com/1Password/terraform-provider-onepassword/issues/312

### ✅ Checklist

- [x] 🖊️ Commits are signed
  - [ ] 🔹 Unit /🔸 Integration
  - [ ] 🌐 E2E

Test with these configs:

```
resource "onepassword_item" "password_recipe_login" {
  vault    = data.onepassword_vault.testing.uuid
  title    = "Password Recipe Test (Top-Level)"
  category = "login"
  username = "recipe_user"

  password_recipe {
    length  = 20
    digits  = true
    symbols = true
  }
}


resource "onepassword_item" "password_recipe_section_list" {
  vault    = data.onepassword_vault.testing.uuid
  title    = "Password Recipe Test (Section List)"
  category = "login"
  username = "recipe_user"

  section {
    label = "Credentials"

    field {
      label = "API Key"
      type  = "CONCEALED"

      password_recipe {
        length  = 20
        digits  = false
        symbols = true
      }
    }
  }
}
```

Import the items and run terraform apply. Confirm the password recipe is just added to state. The password should not be changed.

### DIFF
--- a/internal/provider/value_modifier.go
+++ b/internal/provider/value_modifier.go
@@ -54,7 +54,7 @@ func (m valueModifier) PlanModifyString(ctx context.Context, req planmodifier.St
 	}
 }
 
-// shouldPreservePasswordValue returns true when the modifier should copy state password to plan:
+// shouldPreservePasswordValue returns true when the modifier should copy state password to plan
 func shouldPreservePasswordValue(stateRecipe, planRecipe []PasswordRecipeModel) bool {
 	if reflect.DeepEqual(stateRecipe, planRecipe) {
 		return true


README.md
```diff
@@ -83,7 +83,7 @@ resource "onepassword_item" "demo_login" {
 
 For the contribution guidelines, see [CONTRIBUTING.md](/CONTRIBUTING.md).
 
-Still not sure where or how to begin? We're happy to help! You can join the [Developer Slack workspace](https://join.slack.com/t/1password-devs/shared_invite/zt-1halo11ps-6o9pEv96xZ3LtX_VE0fJQA), and ask us any questions there.
+Still not sure where or how to begin? We're happy to help! You can join the [Developer Slack workspace](https://developer.1password.com/joinslack), and ask us any questions there.
```
JillRegan, line 87: Noticed there was another link in the README

internal/provider/value_modifier.go
```diff
@@ -49,11 +49,27 @@ func (m valueModifier) PlanModifyString(ctx context.Context, req planmodifier.St
...
+func shouldPreservePasswordValue(stateRecipe, planRecipe []PasswordRecipeModel) bool {
+	if reflect.DeepEqual(stateRecipe, planRecipe) {
+		return true
+	}
+
+	stateHasNoRecipe := len(stateRecipe) == 0
+	planHasRecipe := len(planRecipe) > 0
+
+	// If the state has no recipe and the plan has a recipe then the password should be preserved.
+	if stateHasNoRecipe && planHasRecipe {
```
volodymyrZotov, line 67: That won't work for the next case:
1. Create an item with password recipe. (terraform apply)
2. remove item from state `terraform state rm onepassword_item.example`
3. Update password recipe in config and terraform apply again.
4. It won't generate new password but keep initial.

PR #364, opened 2026-04-23
Add onepassword_items data source for batch-reading items

### LINKED ISSUE #363: Feature request: onepassword_items data source for batch-reading multiple items
### Problem

Reading multiple items from a vault requires a separate `data "onepassword_item"` block for each one. In modules that pull 10-20 secrets this creates verbose configs and hits rate limits at scale (related: #184, #343).

```hcl
# Today: one block per item
data "onepassword_item" "db_password" {
  vault = var.vault_id
  title = "DB_PASSWORD"
}
data "onepassword_item" "api_key" {
  vault = var.vault_id
  title = "API_KEY"
}
# ... repeated for every secret
```

### Use cases

Modules that manage infrastructure requiring multiple secrets (databases, APIs, service credentials) need to fetch 10-20+ items from 1Password. The current approach creates verbose Terraform configs and can hit rate limits when modules are instantiated at scale across environments.

### Proposed solution

A new `onepassword_items` (plural) data source that accepts a vault and a list of titles, returning all matched items in a single read.

```hcl
data "onepassword_items" "secrets" {
  vault  = var.vault_id
  titles = ["DB_PASSWORD", "API_KEY", "OAUTH_SECRET"]
}

# Access individual items
locals {
  db_host = data.onepassword_items.secrets.items["DB_PASSWORD"].username
  db_pass = data.onepassword_items.secrets.items["DB_PASSWORD"].password
  api_key = data.onepassword_items.secrets.items["API_KEY"].credential
}
```

The `items` output is a map from title to item attributes (id, title, category, username, password, credential, note_value, url, tags).

Yes - using individual `data "onepassword_item"` blocks for each secret (shown in the "Today" example above), but this is verbose and hits rate limits at scale.

### Implementation notes

I have a working implementation covering both backends:

- **SDK backend**: Resolves titles to IDs via a single `Items().List()` call, then fetches all items in one `Items().GetAll()` batch request. Two API calls total regardless of item count.
- **Connect backend**: Falls back to sequential `GetItem()` calls since Connect has no batch API.
- Adds `GetItems(ctx, vaultUUID, titlesOrIDs)` to the `Client` interface.
- Titles that match zero or multiple items return a clear error.

Happy to open a PR if there's interest. A few design questions I'd want input on:

1. **`titles` input accepting UUIDs too** -- currently the `titles` field accepts both titles and UUIDs (detected via format). Should these be separate attributes instead?
2. **`credentials` convenience map** -- the implementation includes a `credentials` map (title to primary secret, preferring API credential over password). Is this too opinionated for the provider, or useful enough to keep?
3. **Scope** -- the current implementation exposes top-level fields only (no sections or file attachments). Is that a reasonable first pass?

### References & Prior Work

* #184 - Rate limiting issues with multiple item reads
* #343 - Related performance concerns

trgoodwin:
## Summary

Adds a new `onepassword_items` (plural) data source that reads multiple items from a vault in a single Terraform data block. This reduces configuration boilerplate and API calls when modules need 10-20+ secrets.

Closes #363

## Details

- **SDK backend**: Resolves titles via `Items().List()`, then batch-fetches with `Items().GetAll()` -- two API calls total regardless of item count
- **Connect backend**: Falls back to sequential `GetItem()` calls (Connect has no batch API)
- Adds `GetItems(ctx, vaultUUID, titlesOrIDs)` to the `Client` interface
- `titles` field accepts both titles and UUIDs (detected via format)
- Includes a `credentials` convenience map (title -> primary secret value)
- Exposes top-level fields only (no sections or file attachments) as a first pass

## Usage

```hcl
data "onepassword_items" "secrets" {
  vault  = var.vault_id
  titles = ["DB_PASSWORD", "API_KEY", "OAUTH_SECRET"]
}

output "db_password" {
  value     = data.onepassword_items.secrets.credentials["DB_PASSWORD"]
  sensitive = true
}
```

## Open questions

1. Should `titles` accepting UUIDs be split into separate attributes?
2. Is the `credentials` convenience map too opinionated, or useful enough to keep?
3. Is top-level fields only a reasonable scope for v1?

### DIFF
--- a/docs/data-sources/items.md
+++ b/docs/data-sources/items.md
@@ -1,0 +1,58 @@
+---
+page_title: "onepassword_items Data Source - onepassword"
+subcategory: ""
+description: |-
+  Use this to batch-read multiple items from a vault. Returns item details and a convenience credentials map.
+---
+
+# onepassword_items (Data Source)
+
+Use this to batch-read multiple items from a vault by their titles or UUIDs. This is more efficient than using multiple `onepassword_item` data sources with `for_each`, as it reduces Terraform configuration boilerplate and provides a convenient `credentials` map for the common case of reading secret values.
+
+## Example Usage
+
+```terraform
+data "onepassword_items" "secrets" {
+  vault  = "your-vault-id"
+  titles = ["DB_PASSWORD", "API_KEY", "JWT_SECRET"]
+}
+
+# Access individual credentials
+output "db_password" {
+  value     = data.onepassword_items.secrets.credentials["DB_PASSWORD"]
+  sensitive = true
+}
+
+# Access full item details
+output "api_key_category" {
+  value = data.onepassword_items.secrets.items["API_KEY"].category
+}
+```
+
+## Schema
+
+### Required
+
+- `vault` (String) The UUID of the vault the items are in.
+- `titles` (List of String) A list of item titles (or UUIDs) to retrieve from the vault.
+
+### Read-Only
+
+- `credentials` (Map of String, Sensitive) A map from item title to its primary credential value. For API credential items, this is the `credential` field. For all other items, this is the `password` field.
+- `id` (String) The Terraform resource identifier for this data source.
+- `items` (Attributes Map) A map from item title to item details. (see [below for nested schema](#nestedatt--items))
+
+<a id="nestedatt--items"></a>
+### Nested Schema for `items`
+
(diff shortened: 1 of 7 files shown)


trgoodwin:
apologies - accidentally made off a branch rather than the forked main

PR #365, opened 2026-04-23
Add onepassword_items data source for batch-reading items

### LINKED ISSUE #363: Feature request: onepassword_items data source for batch-reading multiple items

trgoodwin:
### ✨ Summary

Adds a `onepassword_items` (plural) data source that reads multiple items from a vault in a single Terraform data block, cutting configuration boilerplate and API calls for modules that need 10-20+ secrets.

- **SDK backend**: resolves titles via `Items().List()`, then batch-fetches with `Items().GetAll()` — two API calls regardless of item count.
- **Connect backend**: falls back to sequential `GetItem()` calls, as Connect has no batch API.
- Adds `GetItems(ctx, vaultUUID, titlesOrIDs)` to the `Client` interface.
- `titles` accepts both titles and UUIDs, detected by format.
- Exposes a `credentials` convenience map holding each item's primary secret.

```hcl
data "onepassword_items" "secrets" {
  vault  = var.vault_id
  titles = ["DB_PASSWORD", "API_KEY", "OAUTH_SECRET"]
}

output "db_password" {
  value     = data.onepassword_items.secrets.credentials["DB_PASSWORD"]
  sensitive = true
}
```

### 🔗 Resolves:

Closes #363

### ✅ Checklist

- [x] 🖊️ Commits are signed
- [x] 🧪 Tests added/updated
  - [x] 🔹 Unit — `internal/onepassword/sdk/client_test.go` covers the batch path against a fake `ItemsAPI`: title resolution, the zero-match and multiple-match errors, request-order preservation across mixed titles and UUIDs, error attribution, and short, long and unrequested batch responses.
  - [x] 🔸 Integration — seven acceptance tests in `internal/provider`, including the first in that package to exercise Connect's title-lookup branch.
  - [ ] 🌐 E2E — not added; happy to add coverage under `test/e2e` if you'd like it for this data source.
- [x] 📚 Docs updated — `examples/data-sources/onepassword_items/data-source.tf` added and `docs/data-sources/items.md` regenerated with `go generate ./...`. CHANGELOG entry added under Latest → Features.

Rebased onto current `main`; module path updated to `/v3`. `go build ./...`, `go vet ./...` and `make test` all pass locally.

Three design calls, each easy to change if you'd prefer otherwise:

1. **`titles` accepts titles and UUIDs in one attribute.** A UUID is detected by format. This diverges from the singular `onepassword_item` data source, which uses separate `title` and `uuid` attributes with `ExactlyOneOf`. That validator does not translate to a list whose entries could be of either kind, and requiring two parallel lists felt worse than format detection. The known cost is that an item whose title is 26 lowercase alphanumerics is read as a UUID. Happy to split it into `titles` and `uuids` if you'd rather keep the two data sources symmetrical.

2. **`items` and `credentials` are keyed by the exact string supplied in `titles`**, not by the item's own title. Keying by the item's title looked more natural but is not safe: two items in a vault can share a title, and requesting both by their distinct UUIDs would collapse them onto one key with last-write-wins. Keying by the request also means every key is readable straight off the configuration and survives an item being renamed in 1Password. `listvalidator.UniqueValues()` on `titles` keeps the map the same length as the request. `TestAccItemsDataSourceSharedTitle` is the regression test.

3. **The `credentials` map stays, and omits rather than empties.** It covers the common case without callers choosing between `items[...].password` and `items[...].credential` per category. An item with neither field gets no entry at all, so a lookup is a configuration error instead of a silently empty secret. Absent fields on `items` entries are null, following the singular data source's convention. Note that the map is sensitive in whole, so it cannot be used directly in `for_each`; the docs say to iterate `items` and index `credentials` inside.

Scope: top-level fields only, no sections or file attachments. Those add significant schema surface and the batch use case is overwhelmingly flat secrets. `onepassword_item` remains available for full item detail, and sections can be layered on later without breaking this schema.

Two pre-existing behaviours I noticed but deliberately left alone, in case they are news to you: the SDK item listing passes no `ItemListFilter`, so an archived item sharing a title could trip the multiple-match error; and `util.IsValidUUID` is a format heuristic used across the provider, with the 26-character-title caveat above.

Risks are low: this is additive, changes no existing data source or resource behaviour, and the only change to shared code is the new `GetItems` method on the `Client` interface.

### DIFF
--- a/docs/data-sources/items.md
+++ b/docs/data-sources/items.md
@@ -1,0 +1,58 @@
+---
+page_title: "onepassword_items Data Source - onepassword"
+subcategory: ""
+description: |-
+  Use this to batch-read multiple items from a vault. Returns item details and a convenience credentials map.
+---
+
+# onepassword_items (Data Source)
+
+Use this to batch-read multiple items from a vault by their titles or UUIDs. This is more efficient than using multiple `onepassword_item` data sources with `for_each`, as it reduces Terraform configuration boilerplate and provides a convenient `credentials` map for the common case of reading secret values.
+
+## Example Usage
+
+```terraform
+data "onepassword_items" "secrets" {
+  vault  = "your-vault-id"
+  titles = ["DB_PASSWORD", "API_KEY", "JWT_SECRET"]
+}
+
+# Access individual credentials
+output "db_password" {
+  value     = data.onepassword_items.secrets.credentials["DB_PASSWORD"]
+  sensitive = true
+}
+
+# Access full item details
+output "api_key_category" {
+  value = data.onepassword_items.secrets.items["API_KEY"].category
+}
+```
+
+## Schema
+
+### Required
+
+- `vault` (String) The UUID of the vault the items are in.
+- `titles` (List of String) A list of item titles (or UUIDs) to retrieve from the vault.
+
+### Read-Only
+
+- `credentials` (Map of String, Sensitive) A map from item title to its primary credential value. For API credential items, this is the `credential` field. For all other items, this is the `password` field.
+- `id` (String) The Terraform resource identifier for this data source.
+- `items` (Attributes Map) A map from item title to item details. (see [below for nested schema](#nestedatt--items))
+
+<a id="nestedatt--items"></a>
+### Nested Schema for `items`
+
+Read-Only:
+
+- `category` (String) The category of the item.
+- `credential` (String, Sensitive) (Only applies to the API credential category) API credential for this item.
+- `id` (String) The UUID of the item.
+- `note_value` (String, Sensitive) Secure Note value.
+- `password` (String, Sensitive) Password for this item.
+- `tags` (List of String) An array of strings of the tags assigned to the item.
+- `title` (String) The title of the item.
+- `url` (String) The primary URL for the item.
+- `username` (String) Username for this item.
--- a/internal/onepassword/client.go
+++ b/internal/onepassword/client.go
@@ -18,6 +18,7 @@ type Client interface {
 	CreateItem(ctx context.Context, item *model.Item, vaultUuid string) (*model.Item, error)
 	UpdateItem(ctx context.Context, item *model.Item, vaultUuid string) (*model.Item, error)
 	DeleteItem(ctx context.Context, item *model.Item, vaultUuid string) error
+	GetItems(ctx context.Context, vaultUUID string, itemTitlesOrIDs []string) ([]*model.Item, error)
 	GetFileContent(ctx context.Context, file *model.ItemFile, itemUUid, vaultUuid string) ([]byte, error)
 	// GetEnvironmentVariables reads variables from a 1Password Environment. Only supported when using the 1Password SDK (service account or desktop app); not supported with 1Password Connect.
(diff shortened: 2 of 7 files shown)


trgoodwin:
Hey @JillRegan, this one's been sitting a while so I've just rebased it onto current main and brought it in line with the PR template. `go build` and `make test` are both green locally, and I regenerated the docs with `go generate`.

Would you be able to kick off the E2E run with `/ok-to-test sha=ff9f500`? Nothing's run on it yet since the fork workflow needs a maintainer to start it. Also happy to hear if the feature isn't something you want in the provider at all, in which case I'll close it rather than keep it open.

PR #366, opened 2026-04-28
ci: add setup-terraform to build and e2e jobs

JillRegan:
### ✨ Summary

Add the hashicorp/setup-terraform action to the Build and Test workflow and to both jobs in the E2E workflow, matching how Verify Docs already installs Terraform.

Our CI runs Terraform acceptance tests which need a terraform binary. The runners don’t ship Terraform, so the test tooling tried to download it and verify HashiCorp’s release signatures. That check started failing because the HashiCorp GPG key used for those signatures expired. Installing Terraform in the workflow avoids that download and verify path, so tests use a normal CLI install instead.

Build and Test and E2E now pass.

### ✅ Checklist

- [x] 🖊️ Commits are signed
  - [ ] 🔹 Unit /🔸 Integration
  - [ ] 🌐 E2E

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -14,6 +14,11 @@ jobs:
         with:
           go-version-file: go.mod
 
+      - name: Install Terraform
+        uses: hashicorp/setup-terraform@v4
+        with:
+          terraform_wrapper: false
+
       - name: Build
         run: go build -v ./...
 
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -33,6 +33,11 @@ jobs:
         with:
           go-version-file: go.mod
 
+      - name: Install Terraform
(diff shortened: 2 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -14,6 +14,11 @@ jobs:
         with:
           go-version-file: go.mod
 
+      - name: Install Terraform
+        uses: hashicorp/setup-terraform@v4
+        with:
+          terraform_wrapper: false
```
bertrmz, line 20: For other reviewers interested in this setting, there's a useful (though slightly outdated) [DeepWiki page](https://deepwiki.com/hashicorp/setup-terraform/3.2-terraform-wrapper) describing the wrapper design.

bertrmz (approved):

rishiy15 (approved):

PR #393, opened 2026-09-01
Expose item created_at and updated_at

paul-robertson-ag:
### ✨ Summary

The 1Password SDK and Connect both report an item's createdAt/updatedAt, but the provider dropped them during conversion, so there was no way to tell when an item was last changed in 1Password.

Carry both through model.Item and expose them as computed RFC 3339 attributes (UTC) on the onepassword_item resource, data source and ephemeral resource. A zero time becomes null rather than 0001-01-01T00:00:00Z.

Two things this needed beyond the plumbing:

  - The resource implements ModifyPlan to mark updated_at unknown when a change is planned. Terraform copies a computed attribute's prior state value into the plan, so the applied value would otherwise disagree with the plan and fail with an inconsistent result error. created_at uses UseStateForUnknown since it never changes.

  - Connect answers a create or update with the item as it was before the write - the same reason the version has to be incremented by hand in UpdateItem - so updatedAt in that response is the time of the previous change. Both write paths already re-fetched the propagated item to wait for Connect's sync and then discarded it; they now report it, falling back to the write response if the item does not propagate in time. Nothing downstream depends on the swap: model.Item.Version never reaches Terraform state.

The e2e item resource test captures the timestamps after create and asserts that the update step leaves created_at alone and moves updated_at forward, so a stale write response fails the run rather than passing silently.

### 🔗 Resolves:

I did not see an open issue for this. The general idea here is to be able to use the `updated_at` on the datasource as a parameter into one of the write only attribute resources.

Example: https://search.opentofu.org/provider/hashicorp/aws/latest/docs/resources/ssm_parameter has a `value_wo_version` which could be the updated_at timestamp which would allow secrets to be pulled from 1Password and then stored in AWS secrets manager without storing the value in the state.

Write only attributes require terraform/tofu >= 1.11 and a provider that makes use of them.
- https://opentofu.org/docs/v1.11/language/ephemerality/write-only-attributes/
- https://developer.hashicorp.com/terraform/plugin/sdkv2/resources/write-only-arguments

### ✅ Checklist

- [x] 🖊️ Commits are signed
  - [x] 🔹 Unit /🔸 Integration
  - [ ] 🌐 E2E

- Code was co-authored by claude
    - I did also review it to the best of my ability
    - I did run a plan with this provider and the datasource did return a created at value


CHANGELOG.md
paul-robertson-ag, line 1: I wasnt sure what to do with this file. Should those points be left there? 

It also looks like this info is too verbose and will need to link back to an issue.

Issue #107, opened 2023-11-22
Field within the section is not created if `password_recipe` is specified.


volodymyrZotov:
### Your environment

Terraform Provider Version: 1.2.1

Connect Server Version: 1.7.2

OS: macOS 14.1.1

Terraform Version: 1.4.6

## What happened?
It doesn't create the field within the section if `password_recipe` is specified.

## What did you expect to happen?
A field with a password is created within the section. It should work for both Connect and Service Accounts.

## Steps to reproduce
1. Define the following resource in your `main.tf` file
```
resource "onepassword_item" "demo_sections" {
  vault = "your_vault_id"

  title    = "Demo Terraform Item with Sections"
  category = "login"
  username = "test@example.com"

  section {
    label = "Terraform Section"

    field {
      label = "API_KEY"
      type  = "CONCEALED"
      value = "2Federate2!"
    }

    field {
      label = "HOSTNAME"
      value = "example.com"
    }
  }

  section {
    label = "Terraform Second Section"

    field {
      label = "App Specific Password"
      type  = "CONCEALED"

      password_recipe {
        length  = 40
        symbols = false
      }
    }

    field {
      label = "User"
      value = "demo"
    }
  }
}
```
2. `terraform apply`
3. Check the created item in 1Password Vault.
4. See that there is no filed with the label `App Specific Password` in `Terraform Second Section` section.

## Notes & Logs
This issue happens for Connect and Service Accounts.

PR #397, opened 2026-09-20
feat: developer item categories, SSH key generation, and password recipe variants

dwhoban:
## Summary

Expands the `onepassword_item` resource to the developer-focused item categories, adds provider-side SSH key generation, and introduces memorable and PIN password recipes.

**New categories** (create + update, template-exact field IDs pinned from `op item template get`):

| Category | Top-level attributes |
|---|---|
| `ssh_key` | `ssh_key_type` (`ed25519`\|`rsa`, default `ed25519`), `ssh_key_bits` (rsa, 2048–4096); computed `private_key` (OpenSSH format, sensitive), `public_key`, `fingerprint`, `key_type` |
| `api_credential` | `credential`, `valid_from` (YYYY-MM-DD), `filename` (+ shared `username`) |
| `server` | `admin_console_url`, `admin_console_username`, `admin_console_password` (Admin Console section) |
| `wireless_router` | `network_name`, `server_address`, `wireless_security`, `wireless_password` |
| `software_license` | `license_key`, `version`, `download_link`, `licensed_to`, `registered_email` |

**SSH keys:** generated provider-side (neither SDK offers generation; the CLI does it client-side too) using the existing OpenSSH marshaling code. Keys are never rotated implicitly: the stored PKCS#8 key is converted back from state on updates, and `ssh_key_type`/`ssh_key_bits` force replacement. `ssh_key_type`/`ssh_key_bits`/`fingerprint`/`key_type` are derived on read so imported state matches.

**Recipes:** `password_recipe` gains a `type` discriminator (`random` \| `memorable` \| `pin`, default `random` — existing configurations are unchanged). `random` gains `exclude_characters` (Connect ≥ v1.4.0); `memorable` adds `word_count` (3–15), `separator`, `capitalize`, `word_list`; `pin` reuses `length`. Since Connect cannot generate memorable/PIN passwords server-side, those are generated locally via the SDK's standalone `Secrets.GeneratePassword` util on both auth paths.

**Also:** `SSH_KEY` field type on custom fields; fixed the model→SDK category map omitting `API_CREDENTIAL`; bumped `onepassword-sdk-go` to stable v0.4.1 (zero API-surface delta, picks up the CGO-free build fix).

Category-managed sections (`admin_console`, `customer`, `publisher`) are excluded from the generic `section` state so their fields surface only as top-level attributes — showing them as sections would create permanent drift.

## Resolves

N/A — no tracking issue. (Changelog entries carry `{issue-number}` placeholders for maintainers to fill or drop.)

## Checklist

- [x] All commits are signed
- [x] Tests added/updated:
  - Unit: recipe parsing, real generation for all three recipe kinds, SSH keygen round-trips (PKCS#8 ⇄ OpenSSH), field-type/category maps
  - Integration (mock Connect server, `make test`): 10 new tests — SSH (ed25519 + RSA), API credential, server, wireless router, software license, memorable, PIN
  - E2E (`test/e2e/item_categories_test.go`): SSH lifecycle incl. import-verify and no-rotation on update, RSA, memorable/PIN, create/update for all four new categories — **run locally with desktop-app auth: all passing**
- [x] Documentation updated: schema descriptions regenerated (`go generate` produces no diff), examples added, changelog entry

## Review Notes & Risks

- **No breaking changes** — everything is additive. One behavioral note: items with previously-unknown field types read back typed instead of empty (possible one-time no-op plan diff after upgrade).
- E2E run against a real account found two state bugs that the mock server had masked (unknown `password` after apply for password-less categories; import parity for `ssh_key_type`/`ssh_key_bits`) — both fixed in `5e109c4`, and the mock no longer papers over that class of bug.
- The mock test server change (filling empty password fields instead of appending) slightly alters existing integration-test fixtures' behavior; all existing tests pass.
- E2E for this fork PR will need `/ok-to-test`.


dwhoban:
Validated this change against 1Password/connect#107 ("SSH Keys added via Connect cannot be read by 1Password CLI"). Findings:

**Root cause of #107:** the Connect REST OpenAPI spec defines `SSH_KEY` only as an item *category* — the field-type enum has no SSH key entry at all. The reporter posted the CLI's field serialization (`SSHKEY`) through REST, which Connect stored as an unknown type, making `private_key` invisible to the CLI and unconcealed.

**This provider, per auth path:**

- **Service account / desktop app (SDK): validated empirically, not affected.** I created an `ssh_key` item via this provider against a real account and read it back with `op item get --format json`: the field is normalized to `SSHKEY`, concealed, stored as PKCS#8, fully CLI-readable — byte-for-byte equivalent to a CLI-created item. Notably the account API *synthesizes* `public_key`, `fingerprint`, and `key_type` fields from the private key.
- **Connect: same risk class as #107.** The provider sends field type `SSH_KEY` (the value connect-sdk-go deliberately added in v1.4.0, distinct from the CLI's `SSHKEY`), but since the REST spec has no SSH key field type, this is unspecced territory. The definitive check is the `e2e-connect` CI job (kind + Helm Connect) on this PR — the local mock server can't exercise it.
- **Key generation is client-side by design**, avoiding the other gap noted in #107 (Connect cannot generate keys).

**Follow-up in bfd1dfc:** since the account API synthesizes the public key data, the provider no longer writes an explicit `public_key` field (it produced a duplicate); `public_key`, `fingerprint`, and `key_type` are now derived from the stored private key on read, which also makes reads independent of synthesized fields on any auth path. The `ssh_key_type` description now documents the Connect limitation with a link to the issue.

/cc @twbrowning — the provider-side SSH key creation here may be relevant to your report: through the service-account/desktop path, items come back correctly typed and concealed; the gap you hit is specific to the Connect REST field-type enum.

dwhoban:
Two follow-up fixes from review of the SSH key and recipe work:

- `7eb945d`: `exclude_characters` was dropped from random recipes on the local generation paths (service account/desktop auth, since the generator API has no exclusion support); it is now enforced via bounded rejection sampling, so all auth paths behave consistently. Conflicting exclusion sets (e.g. excluding all digits while `digits = true`) fail with a clear error.
- `d007db1`: changing an item's category forces replacement and `UseStateForUnknown` carried the previous SSH key values into the replacement plan as known values, so the `IsUnknown()`-only clearing left an old private key in non-SSH items' state. Those Computed-only attributes are now nulled outright for non-SSH categories, with a regression test.

The analogous carry-forward exists for the config-settable `password` attribute on password-less categories — that predates this PR and needs a different fix (category-aware value modifier or Create-path clearing), tracked separately in #398.

dwhoban:
Correction to my earlier note on `d007db1`: the leak vector described there (plan modifiers carrying SSH key values across category replacement) turns out not to be reachable in practice. Terraform core plans replacements twice, and the second pass runs against a null prior state, so `UseStateForUnknown`-style modifiers — including this provider's — leave Computed values unknown for the replacement object. Verified empirically on `main` with plan-level tracing; details in #398, now closed as not reproducible. `d007db1` stands as harmless hardening (the attributes are Computed-only, so nulling them outright for non-SSH categories is still correct), and the same tracing confirms real replacement plans deliver unknowns, not carried values, to `Create`.

